PR #21, opened 2021-01-27
Go must be installed

cwiechmann:
Added a note that Gol must be installed with a link to the install page to golang.org


vivekschauhan (approved):

sbolosan (commented):
I'm ok with the additional instruction/hint that Golang must be installed. I'm not sure you need the [ ] . Just say,  Make sure you have Go installed ( without the braces )

PR #24, opened 2021-02-01
Apigov 17450

Rob-Conan:
Cleaned up gitlab-ci file
Added to .csr-profile.json

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,7 +5,7 @@ image: docker:dind
 ############################################################
 include:
   - project: "scurity/gitlabci"
-    ref: APIGOV-17450
+    ref: master
     file: "/.gitlab-ci-config-ssh.yml"
   - project: "scurity/gitlabci"
     ref: master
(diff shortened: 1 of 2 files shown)


jcollins-axway (approved):

kflan-io (approved):

.gitlab-ci.yml
```diff
@@ -1,5 +1,28 @@
 image: docker:dind
 
+############################################################
+# Section for security tool includes
+############################################################
+include:
+  - project: "scurity/gitlabci"
+    ref: master
+    file: "/.gitlab-ci-config-ssh.yml"
```
dfeldick, line 9: The ssh config is in the scurity folder now. I'm not sure if this is the correct place or not. Isn't this for settings common to multiple repos? If so, then wouldn't this yml file be better suited somewhere else? in APIGOV, we have a folder called beano-cicd where we were intending on putting some of the settings that are common to our agent repos. Maybe this would be a better location for the ssh settings, since there are specific references to the sdk?

Just a thought. If you think it is getter suited where it is, then that's OK with me.

Rob-Conan: I didnt realise there was a beano-cicd common gitlab. It would probably be better suited there, I was just trying to find a quick solution and didnt want to go creating a new repo. 
I'm perfectly happy for it to be moved.

dfeldick: OK. If it's OK, I think I'll merge this for now as it is, and then move it when I continue on with the other things that need to be moved out of the SDK repo. This way, it will get the pipeline as it currently stands, moving along.

dfeldick (approved):

PR #30, opened 2021-02-16
Apigov 17561

dfeldick:


kflan-io (approved):

sbolosan (approved):

CODEOWNERS
```diff
@@ -1,7 +1,9 @@
...
-* @dfeldick @sbolosan @jcollins-axway @kflan-io @vivekschauhan
+#* @dfeldick @sbolosan @jcollins-axway @kflan-io @vivekschauhan
+* @Axway/agent-sdk
 
 #.csr-profile.json requires SPOC approval for any modifications
 # @dfeldick - Primary SPOC
 # @vchauhan- Secondary SPOC
-.csr-profile.json @dfeldick @vivekschauhan @mdenihan @rconan @jcollins-axway @sbolosan @kflan-io
+#.csr-profile.json @dfeldick @vivekschauhan @mdenihan @rconan @jcollins-axway @sbolosan @kflan-io
+.csr-profile.json @Axway/agent-sdk
```
jcollins-axway, line 9: Mark and Robert will not be on that list any longer.  Also I though that this was that group Robert sat up in gitlab?

dfeldick: I was going to tell you about that. So, since Robert did gitLAB, I don't think it will work. The permission to commit the csr file needs to be on gitHUB I guess. So, maybe we need a sdk-spoc team or something?

dfeldick: After we verify this all works, I will get rid of the commented out names.

jcollins-axway: They would have to be invited into the org in order to do that

jcollins-axway (approved):

jcollins-axway (approved):

kflan-io (approved):

PR #36, opened 2021-03-03
APIGOV-00000 - fix swagger parse error

jcollins-axway:


vivekschauhan (approved):

sbolosan (commented):
What happens if yaml defs have single quotes? I pulled this from petstore io

kflan-io (approved):

jcollins-axway:
> What happens if yaml defs have single quotes? I pulled this from petstore io

No change in test, just ran through the test. VSC itself updated the file with the quotes atuomatically.

sbolosan (approved):
Thanks for adding the base path support and beefing the func up 🤙

PR #43, opened 2021-03-08
APIGOV-17769 - add check for agentResourceType already a standard generic resource, if not make it one

kflan-io:


pkg/agent/agent.go
```diff
@@ -319,7 +319,12 @@ func getAgentResource() (*apiV1.ResourceInstance, error) {
 
 	response, err := agent.apicClient.ExecuteAPI(coreapi.GET, agentResourceURL, nil, nil)
 	if err != nil {
```
sbolosan, line 321: this is throwing my off...  why not just return nil, err if err!= nil...
then you can take the if agentResource out of the if and make it stand alone

pkg/agent/agent.go
```diff
@@ -319,7 +319,12 @@ func getAgentResource() (*apiV1.ResourceInstance, error) {
...
-		return nil, err
+		// if agentResourceType is already a standard generic agent resource
+		if agentResourceType == agentTypesMap[agent.cfg.AgentType] {
```
sbolosan, line 323: or move this up under line 317?

pkg/agent/agent.go
```diff
@@ -319,7 +319,12 @@ func getAgentResource() (*apiV1.ResourceInstance, error) {
...
+			return nil, err
+		}
+		AgentResourceType = ""
+		return getAgentResource()
```
dfeldick, line 327: Isn't this calling itself?

pkg/agent/agent.go
```diff
@@ -319,7 +319,12 @@ func getAgentResource() (*apiV1.ResourceInstance, error) {
 
 	response, err := agent.apicClient.ExecuteAPI(coreapi.GET, agentResourceURL, nil, nil)
 	if err != nil {
-		return nil, err
+		// if agentResourceType is already a standard generic agent resource
+		if agentResourceType == agentTypesMap[agent.cfg.AgentType] {
+			return nil, err
+		}
+		AgentResourceType = ""
```
dfeldick, line 326: You maybe didn't do this, but to me code gets confusing if you have variables w/ the same name differing in only the leading upper/lower character. It makes it look like a typo.

pkg/agent/agent.go
jcollins-axway: so this is what Kevin and I worked on together.  AWS and Edge set their resource type to AWSDA/TA, EdgeDA/TA.  If only a generic resource was created an error here is hit that the resource was not found.

We wanted a way to say, lets also check if the agent name give is a generic type before throwing that error.  So we set the type to empty and call this method again, returning the generic resource type call.

pkg/agent/agent.go
jcollins-axway: this prevents an infinite loop, if an error is hit trying to get a generic agent we return the error an ddon't continuously try

pkg/agent/agent.go
jcollins-axway: yes, but resetting the resource type so we look for the generic type rather than dataplane specific type

dfeldick: Probably deserves a big, fat comment in the code then. It certainly isn't self-documenting as it is.

jcollins-axway: Agreed

vivekschauhan (approved):

PR #48, opened 2021-03-12
APIGOV-17949 - update resource version

sbolosan:
Set resource version in PUTs for
apiservice
apiservicerevision
apiserviceinstance
consumerinstance

@vivekschauhan @jcollins-axway  I also updated resource version in client.  However, Vivek, jason and I was a part of a meeting with Seby, Chandan, and Ashutosh... talking about removing the xj-axway-agent env var.  Somehow they are seeing this being set multiple times past agent startup...  We are thinking at this point, we can remove this code.  We can discuss.

Along with that, do you think we need to update the following?
service.go
- line 262
- line 330

subscriptionschema.go
-line 170.

In these instances, we are doing a POST before doing a PUT.  I know you said maybe we need to do a GET on the object first...  what are your thoughts?  The above changes are necessary... what about the service.go and subscripitonschema.go

### DIFF
--- a/pkg/apic/apiservice.go
+++ b/pkg/apic/apiservice.go
@@ -41,6 +41,7 @@ func (c *ServiceClient) buildAPIServiceResource(serviceBody *ServiceBody, servic
 }
 
 func (c *ServiceClient) updateAPIServiceResource(apiSvc *v1alpha1.APIService, serviceBody *ServiceBody) {
+	apiSvc.ResourceMeta.Metadata.ResourceVersion = ""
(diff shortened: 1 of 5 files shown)


sbolosan:
@dfeldick @kflan-io @jcollins-axway @vivekschauhan please review

dfeldick (approved):

kflan-io (approved):

jcollins-axway (approved):

vivekschauhan (approved):

PR #55, opened 2021-03-19
APIGOV-17994 - remove call to set attribute

dfeldick:
@jcollins-axway @sbolosan @kflan-io please review


jcollins-axway (approved):

sbolosan (approved):
Let seby know that we addressed this.  They can watch to make sure this isn't showing up anymore...

kflan-io (approved):

PR #59, opened 2021-03-23
APIGOV-17716 - unify logger for discovery and traceability agents

jcollins-axway:


sbolosan (approved):

kflan-io (approved):

pkg/util/log/log.go
```diff
@@ -1,57 +1,123 @@
...
+	"github.com/elastic/beats/v7/libbeat/logp"
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	debugSelector = "apic-agents"
+	traceSelector = "apic-agents-trace"
+)
+
+var isLogP bool
```
dfeldick, line 15: Could have maybe set a default value of either T or F. Then only need to call fn to change value if so desired.

pkg/util/log/log.go
```diff
@@ -1,57 +1,123 @@
...
+
+//SetIsLogP -
+func SetIsLogP() {
```
dfeldick, line 18: Could just have a single fn SetIsLogP(bool). Don't know which is better. In fact, in combination with a default value, you probably never need to even have one of the functions. If the default is true, you might only ever need to set it to false. Or if the default is false, you might only ever need to set it to true.

sbolosan (approved):

pkg/util/log/log.go
jcollins-axway: I orig just wanted Set, but the tests needed it to be unset.  In actual usage unset is not needed

pkg/util/log/log.go
jcollins-axway: golang defaults all bools to false, that is why I only wanted the set function

pkg/util/log/log.go
dfeldick: ok

pkg/util/log/log.go
dfeldick: oh, right.

dfeldick (approved):

PR #62, opened 2021-03-24
APIGOV-18136 - set config team name

sbolosan:
@jcollins-axway @kflan-io @vivekschauhan @dfeldick please review.
This is going under the assumption that local config takes precedence over resource config.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -35,6 +35,10 @@ const (
 	edgeDataplaneType = "Edge"
 )
 
+const (
+	defaultTeam = "Default Team"
+)
+
 // AgentResourceType - Holds the type for agent resource in Central
 var AgentResourceType string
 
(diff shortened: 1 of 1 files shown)


dfeldick (approved):

kflan-io (approved):

kflan-io (approved):

vivekschauhan (approved):

pkg/agent/agent.go
```diff
@@ -557,5 +557,14 @@ func applyResConfigToCentralConfig(cfg *config.CentralConfiguration, resCfgAddit
...
 
-	cfg.TeamName = resCfgTeamName
+	// If config team is not blank, use cfg.TeamName
+	// If config team is blank, check resource team name.  If resource team name is not blank, use resource team name
+	// If config team is blank and resource team name is blank, set to "" and let central handle setting team based on organization
+	if cfg.TeamName == "" {
+		if resCfgTeamName != "" {
+			cfg.TeamName = resCfgTeamName
+		} else {
+			cfg.TeamName = ""
```
jcollins-axway, line 567: This is redundant as we know TeamName is already ""

jcollins-axway (changes requested):

pkg/agent/agent.go
sbolosan: true... thanks!

dfeldick (approved):

dfeldick (approved):

kflan-io (approved):

pkg/agent/agent.go
```diff
@@ -557,5 +557,9 @@ func applyResConfigToCentralConfig(cfg *config.CentralConfiguration, resCfgAddit
 		log.GlobalLoggerConfig.Level(logLevel).Apply()
 	}
 
-	cfg.TeamName = resCfgTeamName
+	// If resource team name is not blank, apply it
+	if resCfgTeamName != "" {
```
jcollins-axway, line 561: I still think you want the cfg.TeamName == "" in this condition.

What would happen if both are set, the resource one would take precedence when it should not.

sbolosan: crap... sorry i got balls in the air again.

kflan-io (approved):

dfeldick (approved):

jcollins-axway (approved):

PR #69, opened 2021-03-31
APIGOV-18221 - fix blacklisting

dfeldick:


pkg/apic/subscriptionmanager.go
```diff
@@ -111,9 +107,14 @@ func (sm *subscriptionManager) processSubscriptions() {
 		case msg, ok := <-sm.receiveChannel:
 			if ok {
 				subscription, _ := msg.(CentralSubscription)
-				sm.preprocessSubscription(&subscription)
-				if subscription.ApicID != "" && subscription.GetRemoteAPIID() != "" {
-					sm.invokeProcessor(subscription)
+				id := subscription.GetID()
+				if !sm.isItemOnBlacklist(id) {
+					sm.addBlacklistItem(id)
+					sm.preprocessSubscription(&subscription)
```
sbolosan, line 113: Should we modify the preprocessSubcription to return err if there is one and check err before possibly invoking the processor or remove the id from the black list?  I'm just thinking if and how this add/remove from the black list could get out of whack.  If its possible...

jcollins-axway: Agreed.  Also the acceptable name for blacklist and whitelist is not blocklist and allowlist.  If you want to be politically correct.

dfeldick: I don't. But, I don't understand how the blist could get out of date as it currently is. I think we do want to "lock" the subscription ASAP before doing anything with it, and this would include preprocessing it. I could understand returning error from the preprocess and then skipping the invokeProcessor. If the preprocessor has an error, there should be no reason to go invoke the processor then.

sbolosan: " I could understand returning error from the preprocess and then skipping the invokeProcessor. If the preprocessor has an error, there should be no reason to go invoke the processor then.'
Ok.. I can live with that

pkg/apic/subscriptionmanager.go
```diff
@@ -110,9 +110,11 @@ func (sm *subscriptionManager) processSubscriptions() {
 				id := subscription.GetID()
 				if !sm.isItemOnBlacklist(id) {
 					sm.addBlacklistItem(id)
-					sm.preprocessSubscription(&subscription)
-					if subscription.ApicID != "" && subscription.GetRemoteAPIID() != "" {
-						sm.invokeProcessor(subscription)
+					err := sm.preprocessSubscription(&subscription)
+					if err != nil {
+						if subscription.ApicID != "" && subscription.GetRemoteAPIID() != "" {
```
jcollins-axway, line 115: can't these conditions all be combined?

jcollins-axway (changes requested):

pkg/apic/subscriptionmanager.go
dfeldick: yes

vivekschauhan (approved):

kflan-io (approved):

sbolosan (approved):
Leave it for now. If anyone is in this code in the future maybe change it from lock to locked 🔒

PR #74, opened 2021-04-08
APIGOV-18296 - clean out expiry token.  Fetch new

sbolosan:
Clean out expiry token.
then when get cached token is empty, try and fetch new token

### DIFF
--- a/pkg/apic/auth/apicauth.go
+++ b/pkg/apic/auth/apicauth.go
@@ -355,6 +355,8 @@ func (th *tokenHolder) getCachedToken() string {
 	if th.tokens != nil {
 		select {
 		case <-th.expiry.C:
+			// cleanup the token on expiry
+			th.tokens = nil
 			return ""
 		default:
 			return th.tokens.AccessToken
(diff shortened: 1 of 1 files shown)


pkg/apic/auth/apicauth.go
```diff
@@ -379,6 +381,7 @@ func (ptp *platformTokenGetter) Close() error {
 
 // fetchNewToken fetches a new token from the platform and updates the token cache.
 func (ptp *platformTokenGetter) fetchNewToken() (string, error) {
+	log.Debug("Get cached token is empty.  Try and fetch a new token")
```
jcollins-axway, line 384: Trace or Debug?

dfeldick (approved):

pkg/apic/auth/apicauth.go
sbolosan: I was hemming and hawing... I'll move it to trace... don't have an issue with that

jcollins-axway (approved):

PR #101, opened 2021-05-13
APIGOV-18557 - hide unnecessary config flags from help

dfeldick:


pkg/cmd/root.go
```diff
@@ -135,7 +135,7 @@ func NewCmd(rootCmd *cobra.Command, exeName, desc string, initConfigHandler Init
 
 // Add the command line properties for the logger and path config
 func (c *agentRootCommand) addBaseProps() {
-	c.props.AddStringPersistentFlag(PathConfigFlag, ".", "Configuration file path for the agent")
+	c.props.AddStringPersistentFlag(PathConfigFlag, ".", "YAML configuration file path for the agent")
```
jcollins-axway, line 138: Is it confusing to say "YAML configuration.." When this is just the path to where the YAML file is, not the path including the YAML filename?

pkg/cmd/properties/properties.go
```diff
@@ -120,6 +121,7 @@ func (p *properties) AddStringSliceProperty(name string, defaultVal []string, de
 		flagName := p.nameToFlagName(name)
 		p.rootCmd.Flags().StringSlice(flagName, defaultVal, description)
 		p.bindOrPanic(name, p.rootCmd.Flags().Lookup(flagName))
+		p.rootCmd.Flags().MarkHidden(flagName)
```
jcollins-axway, line 124: All properties are hidden?

dfeldick: Almost. The help looks like this now:
Flags:
      --envFile string      Path of the file with environment variables to override configuration
  -h, --help                help for discovery_agent
      --pathConfig string   YAML configuration file path for the agent
      --status              Get the status of all the Health Checks
      --synchronize         Run the sync process for the discovery agent
  -v, --version             version for discovery_agent

envFile, pathConfig, status, and synchronize were either creating flags (not props) or were creating persistent properties (don't know what that diff is). So, the methods I changed are not affecting these.

dfeldick: I toyed with passing a bool to hide/not hide to each of the methods, but that would bubble all over. Or, to copy/paste and have different methods completely to hide/not hide. I decided that for now my changes work.

pkg/cmd/root.go
dfeldick: ah. Maybe "Path to the YAML configuration file for the agent"?

jcollins-axway: Path to directory containing YAML...?

dfeldick: ok

PR #107, opened 2021-05-18
APIGOV-18374 - add client timeouts for central, apimanager, apigateway, & output traceability

kflan-io:


pkg/api/client.go
```diff
@@ -23,7 +23,7 @@ const (
 	PUT    string = http.MethodPut
 	DELETE string = http.MethodDelete
 
-	defaultTimeout     = time.Second * 60
+	defaultTimeout     = time.Second * 30
```
jcollins-axway, line 26: Was this meant to change?

pkg/api/client.go
```diff
@@ -78,13 +83,13 @@ func NewClient(cfg config.TLSConfig, proxyURL string) Client {
 }
 
 func getTimeoutFromEnvironment() time.Duration {
-	cfgHTTPClientTimeout := os.Getenv("HTTP_CLIENT_TIMEOUT")
+	cfgHTTPClientTimeout := os.Getenv("CENTRAL_CLIENTTIMEOUT")
```
jcollins-axway, line 86: Was this supposed to change? CENTRAL_CLIENTTIMEOUT is for central specifically to override what the HTTP_CLIENT_TIMEOUT might have

pkg/api/client.go
kflan-io: Yes; the story asks for 30 second defaults. I contemplated adding a new variable.

jcollins-axway: This is good then.

pkg/api/client.go
sbolosan: following...

pkg/api/client.go
sbolosan: sorry, but did C know that it was at 60 seconds?  I know the story said 30, but I think we moved it to 60 to accommodate the huge petstore swagger suman was testing (if i remember correctlY)

pkg/api/client.go
kflan-io: resolved

jcollins-axway (approved):

sbolosan (approved):

PR #109, opened 2021-05-19
DRAFT APIGOV-18817 Temporary change to allow publishing more than one ConsumerInstance for an API Service

tsjohns9:
I do not plan to merge this branch. It contains a quick hack to allow publishing more than one ConsumerInstance. We will use this branch until the SDK properly supports this featue.

- Hacked a temporary solution to allow publishing of one consumer instance for each API Services. API Service Revisions with the same name and version will fail to publish.
- Fixed a bug with the API Service Instance not having its count incremented, which lead to a 400 error.


pkg/apic/subscription.go
```diff
@@ -36,6 +36,7 @@ type Subscription interface {
 	GetID() string
 	GetName() string
 	GetApicID() string
+	GetRemoteAPIAttributes() map[string]string
```
tsjohns9, line 39: We cannot rely on the `remoteAPIID` attribute any more for mulesoft, because that attribute reflects the API Service only, and not the revision. We are now able to generate traffic and create subscriptions for more than one revision of an API Service, and so we need to find the traffic and the subscriptions based on the api revision id, which we are publishing as the "API ID" attribute.

PR #112, opened 2021-05-21
APIGOV-18817 Small changes to be used by the Mulesoft DA

tsjohns9:
- Export the CentralClient from the agent package.
- Add all API Service attributes to a subscription 
- Fixed a bug with the API Service Instance not having its count incremented, which lead to a 400 error.


pkg/agent/agent.go
```diff
@@ -48,6 +48,8 @@ var agentTypesMap = map[config.AgentType]string{
 	config.TraceabilityAgent: "traceabilityagents",
 }
 
+var CentralClient apic.Client
```
tsjohns9, line 51: Exposing the central client to be used by other agents.

pkg/apic/subscription.go
```diff
@@ -36,6 +36,7 @@ type Subscription interface {
 	GetID() string
 	GetName() string
 	GetApicID() string
+	GetRemoteAPIAttributes() map[string]string
```
tsjohns9, line 39: Adding all the attributes from the API Service to the subscription because the mulesoft agent does not rely on the remoteAPIID attribute to determine anything. It needs other attributes to deal with the specific revision.

pkg/apic/apiservicerevision.go
```diff
@@ -95,6 +95,11 @@ func (c *ServiceClient) processRevision(serviceBody *ServiceBody) error {
 	return nil
 }
 
+// GetAPIRevisions - Returns the list of API revisions for the specified filter
+func (c *ServiceClient) GetAPIRevisions(queryParams map[string]string, stage string) ([]v1alpha1.APIServiceRevision, error) {
```
tsjohns9, line 99: Exposing this func so the DA can retrieve all revisions of an api service at startup for caching.

pkg/agent/agent.go
vivekschauhan: There is already a method GetCentralClient() on agent to get the API client, so this is not needed

vivekschauhan (approved):

jcollins-axway (approved):

kflan-io (approved):

PR #115, opened 2021-05-25
APIGOV-18453 - unhealthy status when job or health check fail

jcollins-axway:


sbolosan (approved):
just curious... i mean i like the stop/start Log func... but it didn't really add anything right?

jcollins-axway:
> just curious... i mean i like the stop/start Log func... but it didn't really add anything right?

It did. The detached interval job is a different jobType. By using those common funcs I did not have to duplicate all the logic in the start/stop funcs of an interval job to the detached interval job.

sbolosan:
> > just curious... i mean i like the stop/start Log func... but it didn't really add anything right?
> 
> It did. The detached interval job is a different jobType. By using those common funcs I did not have to duplicate all the logic in the start/stop funcs of an interval job to the detached interval job.

Ah...  i see it now... I like it.

pkg/agent/periodicstatusupdate.go
```diff
@@ -20,6 +22,12 @@ func (su *periodicStatusUpdate) Ready() bool {
 	if runStatusUpdateCheck() != nil {
 		return false
 	}
+	// Do not start until status will be running// get the status from the health check and jobs
```
dfeldick, line 25: what happened here? 2 lines of comments get merged into 1 line accidentally?

jcollins-axway: looks like it, will fix

dfeldick (approved):

vivekschauhan (approved):

tsjohns9 (approved):

PR #117, opened 2021-05-26
APIGOV-18997 - additional publish methods to pubsub

jcollins-axway:


dfeldick (approved):

sbolosan (approved):
Looks good... thanks!

pkg/cache/pubsub.go
```diff
@@ -118,54 +140,55 @@ func (c *cachePubSub) PublishToTopicWithSecondaryKey(secondaryKey string, data i
...
-func (c *cachePubSub) PublishCacheHash(key, secondarykey string, data interface{}) error {
+func (c *cachePubSub) PublishCacheHash(key, secondaryKey string, data interface{}) error {
 	hash, err := util.ComputeHash(data)
+	if err != nil {
+		return err
+	}
+
 	changed, err := c.hasChanged(key, hash)
 	if !changed || err != nil {
 		return err
```
dfeldick, line 155: could probably still combine all of these lines into a func that returns err, hash, and changed (maybe don't even need changed) because these lines are all repeated multiple x.

jcollins-axway: They are repeated because I don't want to update the cache, or send it on the channel, if the item has not changed.  I combined what I could to get this, but still would require a return and if check if I moved this in anyway.

dfeldick: ok

PR #121, opened 2021-05-27
Apigov api revision

### LINKED ISSUE #116: Update the SDK to use friendly names for the API Service Revisions instead of using GUIDS for the name
Currently there is no way to override the uuid generation for the API revision name. We need to provide an optional setter for the revision that when present will prevent the uuid from being generated.

chrisccoy:
Fixes #116

### DIFF
--- a/pkg/apic/definitions.go
+++ b/pkg/apic/definitions.go
@@ -104,7 +104,7 @@ type ServiceBody struct {
 	ImageContentType  string
 	CreatedBy         string
 	ResourceType      string
-	AltRevisionPrefix    string
+	AltRevisionPrefix string
 	SubscriptionName  string
 	APIUpdateSeverity string `json:",omitempty"`
 	State             string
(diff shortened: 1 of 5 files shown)


sbolosan (approved):

pkg/apic/definitions.go
```diff
@@ -104,6 +104,7 @@ type ServiceBody struct {
 	ImageContentType  string
 	CreatedBy         string
 	ResourceType      string
+	AltRevisionPrefix    string
```
dfeldick, line 107: There is a Prettier code formatter extension (if you are using VSCode) that will auto align things in structs, fix inconsistent whitespacing (which I see elsewhere) that we all have installed.

tsjohns9 (approved):

pkg/apic/definitions.go
chrisccoy: @dfeldick can you send me what your are using and the configuration/setup?

chrisccoy: 👍  code has been formatted

vivekschauhan (approved):

dfeldick (approved):

PR #122, opened 2021-05-27
APIGOV-18499 - add URL validations

dfeldick:


pkg/config/centralconfig.go
```diff
@@ -488,6 +488,17 @@ func (c *CentralConfiguration) validateConfig() {
 	}
 }
 
+func (c *CentralConfiguration) validateURL(urlString, configPath string, isURLRequired bool) {
```
sbolosan, line 491: I guess I don't understand the boolean?  when is the URL not required?  If its not, then why even call validateURL?

dfeldick: A few urls, like those for Proxys, are not required in the config. But, if they are non-blank, they should be validated.

sbolosan (approved):

kflan-io (approved):

jcollins-axway (approved):

PR #128, opened 2021-06-01
APIGOV-18029 - Changes to use secret reference in config

vivekschauhan:
- Changes to secure string based configuration by using reference of secret resource in Central
- Secret resolver gets hooked by in property parser for string flags and property
- Secret resolver parses the secret and key from the reference string value in config and gets the secret from API server
- The secret is cached to minimize the number of API calls to API server to fetch secret in case same secret is used for multiple config
- Also includes a fix to setup beats logger to allow logging any log entries during agent initialization

### DIFF
--- a/pkg/cmd/properties/properties.go
+++ b/pkg/cmd/properties/properties.go
@@ -11,11 +11,20 @@ import (
 	"time"
 
 	"github.com/Axway/agent-sdk/pkg/util"
+	"github.com/Axway/agent-sdk/pkg/util/errors"
 	"github.com/Axway/agent-sdk/pkg/util/log"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
 )
+
+// ErrInvalidSecretReference - Error for parsing properties with secret reference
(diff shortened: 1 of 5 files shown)


pkg/cmd/properties/properties.go
```diff
@@ -11,12 +11,21 @@ import (
...
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
 )
 
+// ErrInvalidSecretReference - Error for parsing properties with secret reference
+var ErrInvalidSecretReference = errors.Newf(1411, "invalid secret reference - %s, please check the value for %s config")
+
+// SecretPropertyResover - interface for resolving property values with secret references
+type SecretPropertyResover interface {
```
jcollins-axway, line 25: Resolver?

jcollins-axway (approved):

pkg/cmd/properties/properties.go
vivekschauhan: Done

sbolosan (approved):

PR #135, opened 2021-06-09
APIGOV-19081 - take advantage of pagination for instance and revision

sbolosan:
Update get instance and get revisions using pagination.


sbolosan:
For apiservicerevision, I kept the signature of the function for queryParams since it looks like mulesoft needs the GetAPIRevisions to still exist in the Client interface.

jcollins-axway:
So the paging is going to be a common implementation for each resource.  Anyway to make it more of a common method that uses the ResourceDefinition struct or an interface and then the caller can convert that to its expected resource type?

sbolosan:
@jcollins-axway , I can get with that!

pkg/apic/service.go
```diff
@@ -25,6 +25,13 @@ const (
 	deleteAPI            = iota
 )
 
+const (
+	apiServerPageSize    = 20
+	healthcheckEndpoint  = "central"
+	attributesQueryParam = "attributes."
+	apiServerFields      = "name,title,attributes"
+)
+
```
dfeldick, line 34: not sure why these showed up here. I only see the pagesize referenced in any of your changes. I also see these exact defs in discoverycache.go. Seems like we shouldn't have an exact copy of them in 2 places.

sbolosan: yeah, will update, i forgot to remove.

dfeldick (approved):

pkg/apic/resourcePagination.go
```diff
@@ -0,0 +1,62 @@
...
+	"encoding/json"
+	"strconv"
+
+	coreapi "github.com/Axway/agent-sdk/pkg/api"
+	"github.com/Axway/agent-sdk/pkg/apic/apiserver/clients/definitions/v1alpha1"
+	"github.com/Axway/agent-sdk/pkg/apic/apiserver/models/management/v1alpha1"
+	"github.com/Axway/agent-sdk/pkg/util/log"
+)
+
+func (c *ServiceClient) fubar() error {
```
dfeldick, line 13: Don't think you wanted to leave this...

pkg/apic/resourcePagination.go
```diff
@@ -0,0 +1,62 @@
...
+func (c *ServiceClient) getAPIResources(queryParams map[string]string, URL, stage string) ([]interface{}, error) {
+	morePages := true
+	page := 1
+
+	resourceInstance := make([]interface{}, 0)
+
+	for morePages {
+		query := map[string]string{
+			"page":     strconv.Itoa(page),
+			"pageSize": strconv.Itoa(apiServerPageSize),
```
dfeldick, line 35: might there ever be cases where you want the pagesize for different resources to be different? If so maybe pass it in rather than having a single global value?

pkg/apic/resourcePagination.go
sbolosan: *NOTE*  - sorry (guys), this was a DRAFT for a single point of pagination func that was supposed to be checked into a different branch. PLEASE ignore for now.

pkg/apic/resourcePagination.go
sbolosan: see above... check in was a DRAFT holder... Sorry for the confusion

pkg/apic/resourcePagination.go
```diff
@@ -5,29 +5,33 @@ import (
...
 
 	coreapi "github.com/Axway/agent-sdk/pkg/api"
-	"github.com/Axway/agent-sdk/pkg/apic/apiserver/clients/definitions/v1alpha1"
+	apiv1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
 	"github.com/Axway/agent-sdk/pkg/apic/apiserver/models/management/v1alpha1"
 	"github.com/Axway/agent-sdk/pkg/util/log"
 )
 
-func (c *ServiceClient) fubar() error {
+func (c *ServiceClient) GetAPIServiceInstances() ([]*v1alpha1.APIServiceInstance, error) {
```
dfeldick, line 13: lint is gonna bit you if you don't put a comment here for the method name

dfeldick (approved):

pkg/apic/apiservicerevision.go
```diff
@@ -107,47 +107,54 @@ func (c *ServiceClient) GetAPIRevisions(queryParams map[string]string, stage str
...
 
-	response, err := c.apiClient.Send(request)
-	if err != nil {
-		return nil, err
-	}
-	if response.Code != http.StatusOK {
-		if response.Code != http.StatusNotFound {
-			responseErr := readResponseErrors(response.Code, response.Body)
-			return nil, utilerrors.Wrap(ErrRequestQuery, responseErr)
+	for morePages {
```
jcollins-axway, line 117: Should this be using the getResources func?

Also, can getResources be made public?  Then the pagination stuff in agent.go may utilize it as well?

sbolosan:
1. Kept GetAPIRevisions in apiservicerevisions.go //TODO - SHANE need to check with mulesoft for this. @tsjohns9 
2. Need to create secondary TECH story for adding item to APICache.
3. Goes with #2. Need to take advantage of pagination for updateAPICache()

@kflan-io @jcollins-axway @dfeldick @vivekschauhan . Should be ready to review.  Please keep in mind the above notes.  Tested and seems to be working, but will have to test against AWS and Azure today.

sbolosan:
`apiServices := make([]apiV1.ResourceInstance, 0)
		json.Unmarshal(response, &apiServices)

		for _, apiService := range apiServices {
			externalAPIID := addItemToAPICache(apiService)
			if externalAPIPrimaryKey, found := apiService.Attributes[apic.AttrExternalAPIPrimaryKey]; found {
				existingAPIs[externalAPIPrimaryKey] = true
			} else {
				existingAPIs[externalAPIID] = true
			}
		}`

@vivekschauhan @jcollins-axway we can take this one off line, but this is the area I did not take advantage of the pagination and thinking I need to spike and do another TECH story...

The new func returns []*apiv1.ResourceInstance

And that's ok, but addItemToAPICache takes apiV1.ResourceInstance.  As I started to go down the rabbit hole, I found around like 10 or so touch points that is either calling this method or using the item id... 

like here...
`// PublishAPI - Publishes the API
func PublishAPI(serviceBody apic.ServiceBody) error {
	if agent.apicClient != nil {
		ret, err := agent.apicClient.PublishService(serviceBody)
		if err == nil {
			apiSvc, e := ret.AsInstance()
			if e == nil {
				addItemToAPICache(*apiSvc)
				//update the local activity timestamp for the event to compare against
				UpdateLocalActivityTime()
			}
		} else {
			return err
		}
	}
	return nil
}`

dfeldick (approved):

sbolosan:
@jcollins-axway @vivekschauhan disregard...
Jason found the one * pointer i was missing and that fixed everything... onward with code review please.

dfeldick (approved):

pkg/apic/apiservicerevision.go
```diff
@@ -102,52 +102,60 @@ func (c *ServiceClient) processRevision(serviceBody *ServiceBody) error {
...
```
jcollins-axway, line 118: This is not supposed to use GetAPIServiceRevisions or similar?

sbolosan: No @jcollins-axway  I did not change this specifically.  Left this one in there for @tsjohns9 to look at.  Because the calling func is coming from Mulesoft I'm assuming since @tsjohns9 created a GetAPIRevisions func

`// GetAPIRevisions - Returns the list of API revisions for the specified filter
func (c *ServiceClient) GetAPIRevisions(queryParams map[string]string, stage string) ([]v1alpha1.APIServiceRevision, error) {
	return c.getAPIRevisions(queryParams, stage)
}`

I tagged him to look at it and see if he can leverage the func in paginationResource like the others are

dfeldick (approved):

jcollins-axway (approved):

kflan-io (approved):

vivekschauhan (approved):

PR #136, opened 2021-06-09
APIGOV-19048 - add healthchecking for condor connections

dfeldick:


kflan-io (approved):

jcollins-axway:
I am confused, are you not registering the health check any longer? Just creating it as a job that is called health check and uses some health check constants?

dfeldick:
Yes. Why would I need both a job and a registration of a healthchecker?

sbolosan:
I think he's saying you need to register the actual job...

dfeldick:
jobs.RegisterIntervalJob. Line 247.

sbolosan:
hmmm...looks like you're doing that

pkg/traceability/traceability.go
```diff
@@ -211,3 +230,78 @@ func updateEvent(batch publisher.Batch, events []publisher.Event) {
...
+			protocol: config.Protocol,
+			host:     config.Hosts[i],
+			proxyURL: config.Proxy.URL,
+			timeout:  config.Timeout,
+		}
+
+		hcJob := &condorHealthCheckJob{
+			agentHealthChecker: ta,
+		}
+		jobs.RegisterIntervalJob(hcJob, config.Timeout)
```
sbolosan, line 247: This can return an error.  You probably should check for that

dfeldick: I can and will. But if I dig into RegisterIntervalJob, it looks like it never really returns anything except nil.

jcollins-axway:
> jobs.RegisterIntervalJob. Line 247.

That works. But then you should not run the function as if its a health check function with health check returns.  All that health check code could be in your job struct.  If you are not registering a healthcheck at all then don't use the healthcheck library at all, is wha tI mean.

dfeldick:
I apparently mistakenly patterned my implementation after pollUsagesPlans in aws, where it does almost the exact same thing I did. However, I see what you are saying I think. I can move the healthcheck stuff from traceability.go directly into my job file and tweek it to just return success/fail or whatever, thereby eliminating any of the hc calls and vars.

jcollins-axway:
> I apparently mistakenly patterned my implementation after pollUsagesPlans in aws, where it does almost the exact same thing I did. However, I see what you are saying I think. I can move the healthcheck stuff from traceability.go directly into my job file and tweek it to just return success/fail or whatever, thereby eliminating any of the hc calls and vars.

Those ones need updated.

pkg/traceability/traceability.go
```diff
@@ -211,3 +227,82 @@ func updateEvent(batch publisher.Batch, events []publisher.Event) {
...
+		}
+		_, err := jobs.RegisterIntervalJob(hcJob, config.Timeout)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// func (ta *traceabilityAgentHealthChecker) healthcheck(name string) *hc.Status {
```
jcollins-axway, line 252: clean these comments up?

jcollins-axway (approved):

sbolosan (approved):

pkg/traceability/traceability.go
dfeldick: Oops. Done.

PR #139, opened 2021-06-14
Apigov - 19048 - add back temporary "old" healthchecker

dfeldick:
Because the jobs interface does not currently do retries and eventually fail, it does not function as a healthchecker the way the other healthcheckers do... it does not terminate the agent when there is an error. So, in order to temporarily make the agent fail on startup as expected if there is an error, I implemented both a healthchecker Job (the way we want to move to in the future) and an "old-style" healthchecker, which in the future (marked by TODO comments) will be removed.

### DIFF
--- a/pkg/traceability/traceability.go
+++ b/pkg/traceability/traceability.go
@@ -1,6 +1,7 @@
 package traceability
 
 import (
+	"fmt"
 	"net/url"
 	"reflect"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/traceability/traceability.go
```diff
@@ -241,10 +246,47 @@ func registerHealthCheckers(config *Config) error {
...
+		err = registerOldHealthChecker(hcJob, ta.host)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
+
+// TODO: From here down all temporary until Jobs interface finishes full implementation
+func registerOldHealthChecker(hcJob *condorHealthCheckJob, host string) error {
```
sbolosan, line 268: Just being picky here, but can we just name it registerHealthChecker?    the comments are fine and indicate it will be removed in the future.  And your PR general comment is good.  So no need to call it "old".

dfeldick: OK.

sbolosan (approved):

jcollins-axway (approved):

PR #140, opened 2021-06-15
APIGOV-18882 - add APIServiceRevision Pattern for formatting revision title

kflan-io:


pkg/config/centralconfig.go
```diff
@@ -616,6 +629,8 @@ func ParseCentralConfig(props properties.Properties, agentType AgentType) (Centr
 	cfg.URL = props.StringPropertyValue(pathURL)
 	cfg.PlatformURL = props.StringPropertyValue(pathPlatformURL)
 	cfg.APIServerVersion = props.StringPropertyValue(pathAPIServerVersion)
+	//TODO: kf does this need to be here?
```
sbolosan, line 632: left over TODO

pkg/apic/service.go
```diff
@@ -114,21 +117,41 @@ func (c *ServiceClient) UpdateSubscriptionDefinitionPropertiesForCatalogItem(cat
...
-	if serviceBody.Stage != "" {
-		addDescription := fmt.Sprintf("StageName: %s", serviceBody.Stage)
-		if len(serviceBody.Description) > 0 {
-			serviceBody.Description = fmt.Sprintf("%s, %s", serviceBody.Description, addDescription)
-		} else {
-			serviceBody.Description = addDescription
+// updateAPIServiceRevisionTitle - called after APIService gets created or updated.
+// Update title after creating or updating APIService according to the APIServiceRevision Pattern
+func (c *ServiceClient) updateAPIServiceRevisionTitle(serviceBody *ServiceBody) {
+	title := c.cfg.GetAPIServiceRevisionPattern() // "{{APIServiceName}} - {{date:YYYY/MM/DD}} - r {{version}}"
```
sbolosan, line 123: So, why is this func  GetAPIServiceRevisionPattern part of the service client?  why not make it a private func within service.go?

jcollins-axway: Could we actually do this parsing in the central config library, when we read the config and validate.  Then you can create a golang template string there and all you have to do is an execute of the template here.

https://golang.org/pkg/text/template/

sbolosan: @jcollins-axway even better...

pkg/apic/service.go
```diff
@@ -31,12 +33,13 @@ func (c *ServiceClient) PublishService(serviceBody ServiceBody) (*v1alpha1.APISe
...
 	}
-	// Update description title after creating APIService to inlcude the stage name if it exists
-	c.postAPIServiceUpdate(&serviceBody)
+
 	err = c.processRevision(&serviceBody)
 	if err != nil {
 		return nil, err
 	}
+	// Update APIServiceRevision title after creating APIService to inlcude the APIServiceRevision Pattern
+	c.updateAPIServiceRevisionTitle(&serviceBody)
```
vivekschauhan, line 42: I think this should be further up before processing the revision, since the revision will be created with NameToPush setup by agent implementation.

pkg/apic/service.go
```diff
@@ -114,21 +117,41 @@ func (c *ServiceClient) UpdateSubscriptionDefinitionPropertiesForCatalogItem(cat
...
-		addDescription := fmt.Sprintf("StageName: %s", serviceBody.Stage)
-		if len(serviceBody.Description) > 0 {
-			serviceBody.Description = fmt.Sprintf("%s, %s", serviceBody.Description, addDescription)
-		} else {
-			serviceBody.Description = addDescription
+// updateAPIServiceRevisionTitle - called after APIService gets created or updated.
+// Update title after creating or updating APIService according to the APIServiceRevision Pattern
+func (c *ServiceClient) updateAPIServiceRevisionTitle(serviceBody *ServiceBody) {
+	title := c.cfg.GetAPIServiceRevisionPattern() // "{{APIServiceName}} - {{date:YYYY/MM/DD}} - r {{version}}"
+	version := strconv.Itoa(serviceBody.serviceContext.revisionCount + 1)
```
vivekschauhan, line 124: Not quite sure if this is correct, revision and version are different

kflan-io: I took the variable `version` directly from the story {{ version }}. I spoke with Cyrille and have changed the wording to {{revision}}.

dfeldick (approved):

PR #148, opened 2021-06-25
APIGOV-19195 - do panic checking for unimplemented IResourceConfigCallback and IConfigValidator interfaces.

dfeldick:
One note. I didn't create an error in errors.go for the new errors. I didn't know if these should be there, as they are developer errors and not system errors per se. I can add them if you want.

As talked about, and error will be logged indicating that either the IValidatorConfig or IResourceConfig function has not been implemented, but the agent will continue its startup.


pkg/agent/util.go
```diff
@@ -22,11 +25,22 @@ func getTimestamp() v1Time.Time {
...
+func ApplyResourceToConfig(cfg interface{}) error {
+	obj := cfg
+	defer func() {
+		if err := recover(); err != nil {
+			str := fmt.Sprintf("%v", err)
+			if strings.Contains(str, "nil pointer dereference") {
+				log.Errorf("The function 'ApplyResources' for interface IResourceConfigCallback is not implemented in %s.", reflect.TypeOf(obj))
+			}
+		}
+		// }
```
sbolosan, line 41: Comment

dfeldick: Good. I meant to do that, as this was kind of crazy. Thx.

kflan-io (approved):

pkg/agent/util.go
```diff
@@ -22,11 +25,26 @@ func getTimestamp() v1Time.Time {
...
+	obj := cfg
+
+	// This defer func is to catch a possible panic that WILL occur if the cfg object that is passed in embedds the IResourceConfig interface
+	// within its struct, but does NOT implement the ApplyResources method. While it might be that this method really isn't necessary, we will
+	// log an error alerting the user in case it wasn't intentional.
+	defer func() {
+		if err := recover(); err != nil {
+			str := fmt.Sprintf("%v", err)
+			if strings.Contains(str, "nil pointer dereference") {
+				log.Errorf("The function 'ApplyResources' for interface IResourceConfigCallback is not implemented in %s.", reflect.TypeOf(obj))
```
jcollins-axway, line 42: Should this be an error and a message specific for devs or a warn and a message like "ApplyResources not yet implemnted"? Maybe ask Cyrille?

Atraya21 (approved):

pkg/agent/util.go
dfeldick: I'll ask

pkg/agent/util.go
```diff
@@ -22,11 +25,26 @@ func getTimestamp() v1Time.Time {
...
+func ApplyResourceToConfig(cfg interface{}) error {
+	obj := cfg
+
+	// This defer func is to catch a possible panic that WILL occur if the cfg object that is passed in embedds the IResourceConfig interface
+	// within its struct, but does NOT implement the ApplyResources method. While it might be that this method really isn't necessary, we will
+	// log an error alerting the user in case it wasn't intentional.
+	defer func() {
+		if err := recover(); err != nil {
+			str := fmt.Sprintf("%v", err)
+			if strings.Contains(str, "nil pointer dereference") {
```
jcollins-axway, line 41: rather then the  str := fmt.Sprintf can you do err.Error() in the if statement

pkg/config/validator.go
```diff
@@ -1,13 +1,32 @@
...
 // Uses reflection to get the IConfigValidator interface on the config struct or
 // struct variable.
 // Makes call to ValidateCfg method except if the struct variable is of CentralConfig type
 // as the validation for CentralConfig is already done during parseCentralConfig
 func ValidateConfig(cfg interface{}) error {
+	obj := cfg
+	// This defer func is to catch a possible panic that WILL occur if the cfg object that is passed in embedds the IConfigValidator interface
+	// within its struct, but does NOT implement the ValidateCfg method. While it might be that this method really isn't necessary, we will
+	// log an error alerting the user in case it wasn't intentional.
+	defer func() {
```
jcollins-axway, line 21: Can these defer funcs be pulled out into a sep func in the package, they are basically dupes of each other

pkg/agent/util.go
dfeldick: C says to do a warn. I'll also check the SDK docs to be sure that implementing this is documented.

pkg/agent/util.go
dfeldick: Yeah. I had Errorf before and did some changes. Will fix.

pkg/config/validator.go
dfeldick: Yeah. I thought about doing that because Sonar is complaining about the complexity being a little high.. Will fix. Just need to pass the proper msg for each.

pkg/agent/util.go
dfeldick: Actually, can't do your suggestion. recover() does not return an error object. It returns an Interface, which under the covers in this case is just a string. Trying to do err.Error() won't compile. So, the %v is the way to basically do a "toString" on the interface.

jcollins-axway: Ahh, did not realize it returned an interface, good call then.

pkg/config/validator.go
dfeldick: done

sbolosan (approved):

jcollins-axway (approved):

tsjohns9 (approved):

jcollins-axway:
Looks good, update the title to something useful, please.

vivekschauhan (approved):

PR #153, opened 2021-06-29
APIGOV-19451 - updates to code gen for subresources

jcollins-axway:
This is to be merged after this iterations release.

The files changed that are not output by the code gen are:

- pkg/agent/discovery_test.go 
- pkg/apic/apiserver/models/api/v1/resourceinstance.go
- scripts/apiserver/resources.tmpl

For the code gen files I made the following changes.

- Owner *struct -> Owner interface{}
- Icon struct -> Icon interface{}
- FromInstance methods utilize the raw json, if available
- AsInstance marshals the resource then unmarshal to ResourceInstance


dfeldick (approved):

vivekschauhan (approved):

pkg/apic/apiserver/models/api/v1/resourceinstance.go
```diff
@@ -1,13 +1,64 @@
...
+)
+
 // ResourceInstance API Server generic resource structure.
 type ResourceInstance struct {
 	ResourceMeta
-	// GENERATE: The following code has been modified after code generation
-	//  	Owner struct{} `json:"owner"`
-	Owner *struct{} `json:"owner,omitempty"`
+
+	Owner interface{} `json:"owner,omitempty"`
```
tsjohns9, line 12: You mentioned it is modified after code gen. Does that mean we have to change it every time we run the code gen?

jcollins-axway: That was the old codegen but our scripts would handle that.  Now its an interface and there are no additional mods needed for Owner

scripts/apiserver/resources.tmpl
```diff
@@ -40,7 +40,9 @@ func init() {
 type {{.res.kind}} struct {
 	apiv1.ResourceMeta
 	{{range $field, $isNotEmpty := .res.fields}}
-		{{$field | strings.Title}} {{if $isNotEmpty}} {{$.res.kind | strings.ReplaceAll "API" "Api" | strings.ReplaceAll "AWS" "Aws"}}{{$field | strings.Title}} {{else}} struct{} {{end}} `json:"{{$field | strings.ToLower}}"`
+		{{ $type := $field | strings.Title }}
+		{{ $varName := $.res.kind | strings.ReplaceAll "API" "Api" | strings.ReplaceAll "AWS" "Aws" }}
+		{{	$type }} {{if $isNotEmpty}} {{$varName}}{{$type}} {{else}} interface{} {{end}} `json:"{{$field | strings.ToLower}}"`
```
tsjohns9, line 45: Why do we need interface{} instead fo struct{} ?

tsjohns9:
We should test these changes out with the istio agents and make sure everything works

scripts/apiserver/resources.tmpl
jcollins-axway: Correct me if I am wrong, but an empty struct says we do not want to store any of this data.  So when we unmarshal json into an empty struct we lose that data.  If we unmarshal it into an interface we retain the data and then can later cast it into the appropriate structure.

jcollins-axway:
@tsjohns9 - agree, I want to get these in ASAP so QA can run their tests as many times as possible.

PR #154, opened 2021-07-01
[AMPGW-66][APIGOV-19451] Handle sub resources in ResourceInstance and add new auto generated Governance model.

mcollins-axway:
The changes in this code review include the following:-

- Updated to the auto-gen code to support subresources as written by Jason, the plan was to maybe push these ahead of this in APIGOV-19451 but they are not yet pushed hence they are here
- Updates to add support for the GovernanceAgent
- Updates to auto-generated api-server structs run against production api-server (as team Waldorf have pushed model changes to prod last night)


mcollins-axway:
Going to create a new PR after rebase

PR #155, opened 2021-07-06
APIGOV-19258 - cleanup info vs debug messages as described in ticket

dfeldick:
Note: Please look at the jira ticket details for https://jira.axway.com/browse/APIGOV-19258. I added comments in the description (marked in red) and in the comment section on the ticket. There was a lot of ambiguity and inconsistency, and I tried to address as best I could without causing SDK breakage and while maintaining some degree of usefulness of the log messages. Please comment if you feel I have interpreted the ticket incorrectly.


pkg/agent/discoverycache.go
```diff
@@ -165,21 +165,21 @@ func shouldDeleteService(apiID, stage string) bool {
 
 func deleteConsumerInstanceOrService(consumerInstance *v1alpha1.ConsumerInstance, externalAPIID, externalAPIStage string) {
 	if shouldDeleteService(externalAPIID, externalAPIStage) {
-		log.Infof("API no longer exists on the dataplane; deleting the API Service and corresponding catalog item %s from Amplify Central", consumerInstance.Title)
+		log.Debugf("API no longer exists on the dataplane; deleting the API Service and corresponding catalog item %s from Amplify Central", consumerInstance.Title)
```
jcollins-axway, line 168: I know these weren't listed in Cyrille's ticket about Info vs Debug, but a couple in here seem like they may need to be Info still

pkg/util/errors/errors.go
```diff
@@ -5,5 +5,5 @@ var (
 	ErrInitServicesNotReady      = New(1001, "failed to initialize. Services are not ready")
 	ErrTimeoutServicesNotReady   = New(1002, "failed with timeout error.  Services are not ready")
 	ErrPeriodicCheck             = Newf(1003, "%s failed.  Services are not ready")
-	ErrStartingAgentStatusUpdate = New(1004, "error starting %s update")
+	ErrStartingAgentStatusUpdate = Newf(1004, "error starting %s update")
```
jcollins-axway, line 8: good catch

dfeldick: My errorcheck was failing. I mentioned it to Shane. I don't know why it wasn't caught when he put this message in.

pkg/agent/discoverycache.go
dfeldick: I thought the same on a few of these delete-type messages.

jcollins-axway: yeah, when we talked I had not thought of these. Maybe just the first ones in each path? Or the last ones?

dfeldick: I think for consistency with the create messages, I will get rid of "Amplify Central" from the message.

jcollins-axway (approved):

sbolosan (approved):

PR #158, opened 2021-07-09
APIGOV-19572 - report last activity based on activity frequency

sbolosan:
1. Checks are run based on -> CENTRAL_REPORTACTIVITYFREQUENCY
2. When frequency hits, update last activity time for agent resource (and ultimately UI)
3. Only report CENTRAL_REPORTACTIVITYFREQUENCY for periodic checks.  Filter on periodic and not immediate.  Immediate only used for agent status 
4. Refactor time since only periodic status uses it

### DIFF
--- a/pkg/agent/statusupdate.go
+++ b/pkg/agent/statusupdate.go
@@ -72,8 +72,6 @@ func (su *agentStatusUpdate) Execute() error {
 			su.prevStatus = status
 			return nil
 		}
-
-		UpdateLocalActivityTime()
 	}
 
 	// If its a periodic check, tickle last activity so that UI shows agent is still alive.  Not needed for immediate check.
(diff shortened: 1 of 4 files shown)


dfeldick (approved):

jcollins-axway (approved):

pkg/agent/statusupdate.go
```diff
@@ -163,7 +164,3 @@ func runStatusUpdateCheck() error {
 func UpdateLocalActivityTime() {
```
jcollins-axway, line 164: Are we even using the currentActivityTime now? Is this function needed?

sbolosan: You're right...will remove

PR #159, opened 2021-07-09
APIGOV-18722 - add single run job to check for a newer agent version on jfrog

kflan-io:


pkg/agent/agentversionsjob.go
```diff
@@ -0,0 +1,201 @@
...
+	log "github.com/Axway/agent-sdk/pkg/util/log"
+)
+
+var agentURL = map[string]string{
+	"AWSDiscoveryAgent":                      "aws-apigw-discovery-agent",
+	"AWSTraceabilityAgent":                   "aws-apigw-traceability-agent",
+	"AzureDiscoveryAgent":                    "azure-discovery-agent",
+	"AzureTraceabilityAgent":                 "azure-traceability-agent",
+	"EnterpriseEdgeGatewayDiscoveryAgent":    "v7-discovery-agent",
+	"EnterpriseEdgeGatewayTraceabilityAgent": "v7-traceability-agent",
```
tsjohns9, line 26: Should the istio agent be here too?

tsjohns9:
Any tests for the new code?

kflan-io:
I am working on the tests, but wanted to get a review started.

pkg/agent/agentversionsjob.go
kflan-io: The other two options on jfrog are config-sync-agent, and service-mesh-agent. I can map those two build names to those agents if available.

pkg/agent/agentversionsjob.go
```diff
@@ -0,0 +1,201 @@
...
+	//regex check for semantic versioning
+	semVerRegexp := regexp.MustCompile(`\d.\d.\d`)
+	if avj.buildVersion == "" || !semVerRegexp.MatchString(avj.buildVersion) {
+		return errors.ErrStartingVersionChecker.FormatError("build version is missing or of noncompliant semantic versioning")
+	}
+	return nil
+}
+
+// getJFrogVersions - obtaining the versions from JFrog website
+// **Note** this is a temporary solution to obtaining the list of versions by querying jfrog
```
sbolosan, line 98: For this note, can you put a quick spiel on what the desired solution is

pkg/agent/agentversionsjob.go
Atraya21: Those are meshv1 agents. No options for the meshv2 agents? als-traceability-agent or discovery-agent

pkg/agent/agentversionsjob.go
```diff
@@ -0,0 +1,201 @@
...
+		return err
+	}
+	err = avj.getJFrogVersions(avj.urlName)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	// compare build to latest version
+	if isVersionStringOlder(avj.buildVersion, avj.latestVersion) {
+		log.Infof("Running older version of %s. Please consider upgrading from version %s to version %s", avj.dataPlaneType, avj.buildVersion, avj.latestVersion)
```
jcollins-axway, line 78: Should this be a warn?

pkg/agent/agentversionsjob.go
```diff
@@ -0,0 +1,201 @@
...
+	return isVersionSmaller(vB, vL)
+}
+
+// isVersionSmaller - return true if version1 smaller than version2
+func isVersionSmaller(v1 version, v2 version) bool {
+	if v1.major < v2.major {
+		return true
+	} else if v1.major == v2.major && v1.minor < v2.minor {
+		return true
+	} else if v1.major == v2.major && v1.minor == v2.minor && v1.patch < v2.patch {
```
dfeldick, line 127: You could combine this a bit so you don't check major-major 2x:

if v1.major < v2.major {
		return true
} else if v1.major == v2.major {
  if v1.minor < v2.minor {
	return true
  } else if v1.minor == v2.minor && v1.patch < v2.patch {
...

dfeldick: Also, go doesn't generally like all of the if/elseif things. So it would be:

if x < y {
 return true
}
if a == b {
  if c < d {
    return true
  }
  ... etc

pkg/agent/agentversionsjob.go
kflan-io: I chatted with Trevor and opted to add those two in another story, as the location of their docker releases are structured differently. Will add those after they are aligned with these agents for continuity. 
https://jira.axway.com/browse/APIGOV-19578

vivekschauhan: Could this be configured by the agent, so we don't have to keep this mapping

pkg/agent/agentversionsjob.go
```diff
@@ -0,0 +1,207 @@
...
+// Status -
+func (avj *agentVersionCheckJob) Status() error {
+	return nil
+}
+
+// Execute - run agent version check job one time
+func (avj *agentVersionCheckJob) Execute() error {
+	if avj.urlName == "AgentSDK" || avj.urlName == "" {
+		err := errors.ErrStartingVersionChecker.FormatError("empty or generic data plane type name")
+		log.Error(err)
```
vivekschauhan, line 63: Do we want to hide this error or may be trace it, so that non-Axway agent don't keep having this error

kflan-io: I switched to Traces.

jcollins-axway (approved):

pkg/agent/agentversionsjob.go
kflan-io: @vivekschauhan That was the original plan, to send in this info from the agents. However, if this solution does change eventually, I chose to keep all of the related code in one place. I am open to removing the mapping, or making a change part of an APIGOV-19578 solution.

Atraya21 (approved):

PR #165, opened 2021-07-14
APIGOV-19570 - Limiting the calls for subscription schema updates 

vivekschauhan:
- Updates the subscription schema registration logic to use a local cache to identify if the schema requires updates.
- The subscription schema is retrieved on startup using GET call and stored in local cache
- On update lookup is done against the local cached if subscription schema requires update.
- When POST or PUT calls are placed for creating/updating the subscription schema, the schema is added to the local cache

### DIFF
--- a/pkg/apic/consumerinstance.go
+++ b/pkg/apic/consumerinstance.go
@@ -95,9 +95,7 @@ func (c *ServiceClient) enableSubscription(serviceBody *ServiceBody) bool {
 	enableSubscription := serviceBody.AuthPolicy != Passthrough
 	// if there isn't a registered subscription schema, do not enable subscriptions,
 	// or if the status is not PUBLISHED, do not enable subscriptions
-	if enableSubscription && c.RegisteredSubscriptionSchema == nil ||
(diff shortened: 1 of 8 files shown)


sbolosan (approved):

jcollins-axway (approved):

pkg/apic/consumerinstance.go
```diff
@@ -95,9 +95,7 @@ func (c *ServiceClient) enableSubscription(serviceBody *ServiceBody) bool {
 	enableSubscription := serviceBody.AuthPolicy != Passthrough
 	// if there isn't a registered subscription schema, do not enable subscriptions,
```
dfeldick, line 96: I guess this comment isn't quite right now?

kflan-io (approved):

dfeldick (approved):

Atraya21 (approved):

sbolosan (approved):
Night night Vivek.

dfeldick (approved):

jcollins-axway (approved):

tsjohns9 (approved):

PR #166, opened 2021-07-16
APIGOV-19648 - jobs and discovery cache performance changes

jcollins-axway:
- Added ability to set page size when using api server paging methods
- Set page size for discovery cache to 100
- removed query filter on discovery cache and check locally for externalapiid attr
- getting api service instances in place of consumer instances
-- using cache to check for apis removed on dataplane
- only querying for new services and instances, hourly querying for all services and instances
- backoff implemented in jobs api

Note: not using the attribute on the agent resource, replaced with the timestamp query

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -163,21 +163,25 @@ func OnAgentResourceChange(agentResourceChangeHandler ConfigChangeHandler) {
 
 func startAPIServiceCache() {
 	// register the update cache job
-	newDiscoveryCacheJob := newDiscoveryCache(false)
-	id, err := jobs.RegisterIntervalJobWithName(newDiscoveryCacheJob, agent.cfg.GetPollInterval(), "New APIs Cache")
-	if err != nil {
-		log.Errorf("could not start the New APIs cache update job: %v", err.Error())
-		return
-	}
-	log.Tracef("registered API cache update job: %s", id)
-
 	allDiscoveryCacheJob := newDiscoveryCache(true)
-	id, err = jobs.RegisterIntervalJobWithName(allDiscoveryCacheJob, time.Hour, "All APIs Cache")
+	id, err := jobs.RegisterIntervalJobWithName(allDiscoveryCacheJob, time.Hour, "All APIs Cache")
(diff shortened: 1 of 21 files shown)


vivekschauhan (approved):

vivekschauhan (approved):

dfeldick (approved):

kflan-io (approved):

pkg/agent/agent.go
```diff
@@ -161,12 +163,27 @@ func OnAgentResourceChange(agentResourceChangeHandler ConfigChangeHandler) {
...
 	}
-	log.Tracef("registered API cache update job: %s", id)
+	log.Tracef("registered API cache update all job: %s", id)
+
+	// Start the regular update after the first interval
+	go func() {
+		newDiscoveryCacheJob := newDiscoveryCache(false)
+		newDiscoveryCacheJob.lastInstanceTime = time.Now()
+		newDiscoveryCacheJob.lastServiceTime = time.Now()
+		time.Sleep(agent.cfg.GetPollInterval())
```
sbolosan, line 179: Will having this configurable be an issue? Haven't checked, but do we validate this config for < 0 or non numeric

jcollins-axway: Has to be a positive value, > 0

jcollins-axway: Also the cfg library has it set as an int, so that validation is done

sbolosan: Too lazy to check just wanted to make sure they can't do -1

sbolosan (approved):
The change to page size of 100 a seby recommendation?

jcollins-axway:
> The change to page size of 100 a seby recommendation?

Yes, we discussed that since we are limiting the fields to name,title,attributes

pkg/agent/agent.go
jcollins-axway: We, as discussed, should still add a floor, as 1ns is valid....

Atraya21 (approved):

tsjohns9 (approved):

PR #172, opened 2021-07-22
APIGOV-18415 - add ability to redact JMS properties

jcollins-axway:


sbolosan (approved):
liked that you took it out of definitions and created  jmsXXX go files...

pkg/transaction/jmsprotocol.go
```diff
@@ -0,0 +1,86 @@
...
+	return jmsPropertyStringMap, nil
+}
+
+// FromMapStringString - convert the map[string]string to a JMSProtocol
+func (j *JMSProtocol) FromMapStringString(propertyMap map[string]string) error {
+	ignoreFields := map[string]bool{
+		"Type":         true,
+		"JMSStatus":    true,
+		"JMSTimestamp": true,
+	}
```
jcollins-axway, line 54: Should I ignore these here (like I am), expect them to have been set on the object, or ignore these fields in the redaction?

@sbolosan

sbolosan: yeah... To me, it looks ok to be here. Hmmm, yeah, i like it here

dfeldick (approved):

vivekschauhan (approved):

PR #175, opened 2021-07-23
APIGOV-19469 - add SDKversion to agentresources

dfeldick:


sbolosan (approved):

pkg/apic/apiserver/models/management/v1alpha1/Secret.go
```diff
@@ -21,7 +21,7 @@ var (
 )
 
 const (
-	SecretScope = "Environment"
+	SecretScope = "Integration"
```
jcollins-axway, line 24: We probably have to look into scenarios like this as Secrets, and others, can be in multiple scopes

`
secrets                   secret             Secret                          true    Environment        management    
secrets                   secret             Secret                          true    Integration        management 
`

jcollins-axway (approved):

PR #176, opened 2021-07-26
Include json tags for finalizers and remove omitempty

bladedancer:
The lack of finalizers json tags makes them un-editable. The inclusion of omitempty on the maps (attributes, tags, finalizers) makes them add only which isn't what's required.


sbolosan (approved):

vivekschauhan (approved):

dfeldick (approved):

jcollins-axway:
With the removal of omitempty we will have to take care when doing any update to first get the resource and make sure our local cache has the necessary values, similar to the category issue @dfeldick worked on.

PR #180, opened 2021-07-29
APIGOV-19290 - call out env var when template is not valid

jcollins-axway:


sbolosan:
Looks good. Makes very good sense for the error variables.  Easier check 🤙

tsjohns9 (approved):

vivekschauhan (approved):

PR #186, opened 2021-08-04
APIGOV-16307 - add teams to org mapping

dfeldick:


pkg/apic/client.go
```diff
@@ -274,7 +276,23 @@ func (c *ServiceClient) checkAPIServerHealth() error {
...
 	}
-	return nil
+
+	// reset the cache of team names
+	return c.setTeamCache()
+}
+
+func (c *ServiceClient) setTeamCache() error {
+	// passing nil to getTeam will return the full list of teams
+	platformTeams, err := c.getTeam(nil)
```
sbolosan, line 286: Would making it map[string]string{} return a full as well?

dfeldick: probably. I didn't try that one. Should I change it?

pkg/apic/consumerinstance.go
```diff
@@ -32,6 +33,20 @@ func (c *ServiceClient) buildConsumerInstanceSpec(serviceBody *ServiceBody, doc
...
+	owningTeam := c.cfg.GetTeamName()
+
+	// If there is an organizationName in the serviceBody, try to find a match in the map of Central teams.
+	// If found, use that as the owningTeam for the service. Otherwise, use the configured default team.
+	if serviceBody.OrganizationName != "" {
+		obj, err := cache.GetCache().Get(TeamMapKey)
+		if err == nil {
+			teamMap := obj.(map[string]string)
+			if _, found := teamMap[serviceBody.OrganizationName]; found {
+				owningTeam = serviceBody.OrganizationName
```
sbolosan, line 45: Does it buy us anything to log debug if we changed from using default team to use the organizationName?

pkg/apic/client.go
sbolosan: I dunno... what's the stance of sending in nil vs. an instantiated empty?

sbolosan:
Can we edit the title to caps APIGOV and add a description as well...

pkg/apic/consumerinstance.go
dfeldick: Not sure. Depends on what is "expected" behavior. If we expect that they have everything matched up, then maybe logging that we couldn't find a match would be better.

pkg/apic/client.go
dfeldick: probably only matters if the func being called handles them correctly. If it didn't handle nil, it wouldn't have worked and I wouldn't have done it that way.

pkg/apic/servicebuilder.go
```diff
@@ -203,11 +204,17 @@ func (b *serviceBodyBuilder) SetUnstructuredFilename(filename string) ServiceBui
...
 func (b *serviceBodyBuilder) SetAltRevisionPrefix(revisionPrefix string) ServiceBuilder {
 	b.serviceBody.AltRevisionPrefix = revisionPrefix
 	return b
 }
 
+func (b *serviceBodyBuilder) SetOrganizationName(id string) ServiceBuilder {
+	b.serviceBody.OrganizationName = id
+	return b
+}
+
```
jcollins-axway, line 217: Can this be SetTeamName?  In build check if a team name was set (and if the team exists in the cache) otherwise use the default team.  Then the service body you can just use the team name without additional checks?

pkg/apic/client.go
jcollins-axway: I think Shane is right, you can send in the empty map as it would result in no query params just the same.

dfeldick: I will change it, but I've found other examples (e.g. GetProxyList in pollfrontendproxyjob.go, among others) that also are passing in nil for the query params rather than an empty map.

pkg/apic/servicebuilder.go
dfeldick: Yeah. I see now that the var in ServiceBody should be called TeamName also, rather than OrganizationName, as it ultimately becomes OwningTeam in the instanceSpec. So, I'll do some renaming, etc.

dfeldick: Jason's comment about moving the checks proved to not be possible without breaking changes to other agents caused by cyclic imports. After discussing the problem, we decided to leave it pretty much as-is, except for the renaming.

vivekschauhan (approved):

PR #190, opened 2021-08-05
APIGOV-19515 - add team name mismatch error

dfeldick:
add info msg

### DIFF
--- a/pkg/apic/errors.go
+++ b/pkg/apic/errors.go
@@ -31,6 +31,8 @@ var (
 	ErrGetCatalogItemServerInfoProperties = errors.New(1157, "error getting catalog item API server info properties")
 	ErrSubscriptionManagerDown            = errors.New(1158, "subscription manager is not running")
 
-	// Service body builer
+	// Service body builder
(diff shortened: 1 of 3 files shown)


pkg/apic/errors.go
```diff
@@ -31,6 +31,8 @@ var (
 	ErrGetCatalogItemServerInfoProperties = errors.New(1157, "error getting catalog item API server info properties")
 	ErrSubscriptionManagerDown            = errors.New(1158, "subscription manager is not running")
 
-	// Service body builer
+	// Service body builder
 	ErrSetSpecEndPoints = errors.New(1160, "error getting endpoints for the API specification")
+
+	ErrTeamMismatch = errors.Newf(1164, "Amplify Central does not contain a team named %s for API %s. The Catalog Item will be assigned to team %s.")
```
sbolosan, line 37: Mismatch or not found? 😬

jcollins-axway (approved):

pkg/apic/errors.go
dfeldick: Well, I suppose ErrTeamNotFound, but somebody already stole that error name for something else.

PR #191, opened 2021-08-05
APIGOV-19786 - offline usage report generation

jcollins-axway:
Subha wanted to start testing this so we can go through a few things before end of iteration.  I still have unit tests to add.


sbolosan (approved):
Liked that you moved those configs into a config file for grouping purposes...

vivekschauhan (approved):

sbolosan (approved):

kflan-io (approved):

pkg/config/usagereportingconfig.go
```diff
@@ -74,30 +74,36 @@ func NewUsageReporting() UsageReportingConfig {
 		Offline:           false,
 		Schedule:          "@hourly",
 		reportSchedule:    "@monthly",
-		reportGranularity: 3600000,
+		reportGranularity: 900000,
 		qaVars:            false,
 	}
 }
 
+func (u *UsageReportingConfiguration) deprecationWarning(old string, new string) {
```
sbolosan, line 82: Nice... i should go back... when you updated the go template stuff, did you also create a deprecation warning func?  if not, maybe I should go back in and create one.

jcollins-axway:
@vivekschauhan @kflan-io @sbolosan @dfeldick @Atraya21 @sagarkal @tsjohns9 

This is ready for full code review all tests have been added.  Will have one more commit for update env vars.

pkg/config/usagereportingconfig.go
jcollins-axway: I didn't that is a good idea.  I think this should be moved to the logging package? As it can be reused elsewhere.

sbolosan: even better... if you do it, i'll use it... otherwise, i can and move it to the logging package and change this one... either or.

jcollins-axway: Let me see if I can add it here, no probs

Atraya21 (approved):

pkg/config/usagereportingconfig.go
```diff
@@ -0,0 +1,265 @@
...
+// UsingQAVars - Returns the offline boolean
+func (u *UsageReportingConfiguration) UsingQAVars() bool {
+	return u.qaVars
+}
+
+// AddUsageReportingProperties - Adds the command properties needed for Uage Reporting Settings
+func AddUsageReportingProperties(props properties.Properties) {
+	props.AddBoolProperty(pathUsageReportingPublish, true, "Indicates if the agent can publish usage event to Amplify platform. Default to true")
+	props.AddBoolProperty(pathUsageReportingPublishMetric, false, "Indicates if the agent can publish metric event to Amplify platform. Default to false")
+	props.AddDurationProperty(pathUsageReportingInterval, 15*time.Minute, "The time interval at which usage and metric event will be generated")
```
dfeldick, line 247: in these 3 lines, isn't it supposed to be "events", not "event"?

vivekschauhan (approved):

pkg/config/usagereportingconfig.go
jcollins-axway: good call will change, also need to simplify the validate method for Sonar while I am in there

dfeldick (approved):

dfeldick (approved):

PR #194, opened 2021-08-09
APIGOV-19886 - fix init problem causing AgentResource values to not be merged

dfeldick:
1) fix error message if no default team is provided
2) fix init problems causing merged resource values to be overwritten by config values

### DIFF
--- a/pkg/apic/consumerinstance.go
+++ b/pkg/apic/consumerinstance.go
@@ -44,6 +44,11 @@ func (c *ServiceClient) buildConsumerInstanceSpec(serviceBody *ServiceBody, doc 
 			if _, found := teamMap[serviceBody.TeamName]; found {
 				owningTeam = serviceBody.TeamName
 			} else {
+				if owningTeam == "" {
+					owningTeam = "the default team"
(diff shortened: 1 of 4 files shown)


pkg/apic/consumerinstance.go
```diff
@@ -44,6 +44,11 @@ func (c *ServiceClient) buildConsumerInstanceSpec(serviceBody *ServiceBody, doc
 			if _, found := teamMap[serviceBody.TeamName]; found {
 				owningTeam = serviceBody.TeamName
 			} else {
+				if owningTeam == "" {
+					owningTeam = "the default team"
```
sbolosan, line 48: Maybe I'm missing something.  But why not set this as the default?

dfeldick: I guess I can

jcollins-axway (approved):

pkg/apic/consumerinstance.go
sbolosan: Ok, i didn't scroll high enough to see it.

PR #198, opened 2021-08-12
APIGOV-19810 - create concrete struct for Owner rather than using interface

dfeldick:
- Create Owner class and reference from ResourceInstance rather than using interface
- enhance codegen to replace Owner interface in apiserver files with apiv1.Owner (the new struct)
- regen apiserver files

### DIFF
--- a/pkg/apic/apiserver/models/api/v1/resourceinstance.go
+++ b/pkg/apic/apiserver/models/api/v1/resourceinstance.go
@@ -9,7 +9,7 @@ import (
 type ResourceInstance struct {
 	ResourceMeta
 
-	Owner *Owner `json:"owner,omitempty"`
+	Owner *Owner `json:"owner"`
 	// Resource instance specs.
 	Spec map[string]interface{} `json:"spec"`
 
--- a/pkg/apic/apiserver/models/api/v1/resourceinstance.go
+++ b/pkg/apic/apiserver/models/api/v1/resourceinstance.go
@@ -9,7 +9,7 @@ import (
 type ResourceInstance struct {
 	ResourceMeta
 
-	Owner interface{} `json:"owner,omitempty"`
+	Owner *Owner `json:"owner,omitempty"`
 	// Resource instance specs.
 	Spec map[string]interface{} `json:"spec"`
 
--- a/pkg/apic/apiserver/models/catalog/v1alpha1/Asset.go
+++ b/pkg/apic/apiserver/models/catalog/v1alpha1/Asset.go
@@ -40,7 +40,7 @@ type Asset struct {
 
 	Icon interface{} `json:"icon"`
 
-	Owner interface{} `json:"owner"`
+	Owner *apiv1.Owner `json:"owner"`
 
 	References AssetReferences `json:"references"`
 
--- a/scripts/apiserver/resources.tmpl
+++ b/scripts/apiserver/resources.tmpl
@@ -42,7 +42,7 @@ type {{.res.kind}} struct {
 	{{range $field, $isNotEmpty := .res.fields}}
 		{{ $type := $field | strings.Title }}
 		{{ $varName := $.res.kind | strings.ReplaceAll "API" "Api" | strings.ReplaceAll "AWS" "Aws" }}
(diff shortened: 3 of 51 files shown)


pkg/apic/apiserver/models/catalog/v1alpha1/Asset.go
```diff
@@ -40,7 +40,7 @@ type Asset struct {
 
 	Icon interface{} `json:"icon"`
 
-	Owner interface{} `json:"owner"`
+	Owner *apiv1.Owner `json:"owner"`
```
jcollins-axway, line 43: Should these all have omitempty?

dfeldick: I'm wondered the same thing. The first generated code that was blowing up when it wasn't a pointer was Secret.go. Turning it into a pointer fixed it, even without the omitempty. So, it seems like it is fine as is. I wasn't supplying an Owner to Secret, so it would have been empty.

jcollins-axway: Can you check what that payload was?  Would it make sense to send this when the owner info is not sent?

`
{
 "name": "secret",
 "owner": {},
 "spec": ...
}
`

vs

`
{
 "name": "secret",
 "spec": ...
}
`

scripts/apiserver/resources.tmpl
```diff
@@ -42,7 +42,7 @@ type {{.res.kind}} struct {
 	{{range $field, $isNotEmpty := .res.fields}}
 		{{ $type := $field | strings.Title }}
 		{{ $varName := $.res.kind | strings.ReplaceAll "API" "Api" | strings.ReplaceAll "AWS" "Aws" }}
-		{{	$type }} {{if $isNotEmpty}} {{$varName}}{{$type}} {{else}} interface{} {{end}} `json:"{{$field | strings.ToLower}}"`
+        {{ $type}} {{if $isNotEmpty}} {{ $varName }}{{ $type }} {{else}} {{if eq $type "Owner"}} *apiv1.Owner {{else}} interface{} {{end}} {{end}} `json:"{{$field | strings.ToLower}}"`
```
tsjohns9, line 45: Looks like it can be of type Owner or an interface{}. I'm wondering why some of the generated types would have interface{} set instead of Owner for each one?

pkg/apic/apiserver/models/api/v1/resourceinstance.go
```diff
@@ -9,7 +9,7 @@ import (
 type ResourceInstance struct {
 	ResourceMeta
 
-	Owner interface{} `json:"owner,omitempty"`
+	Owner *Owner `json:"owner,omitempty"`
```
jcollins-axway, line 12: If the others are not omitempty, should this be? Should be consistent, no?

tsjohns9 (approved):

pkg/apic/apiserver/models/catalog/v1alpha1/Asset.go
dfeldick: I'm not sure why we would want to treat owner here any different than any of the other possible fields (e.g. attributes, tags, etc) that we don't pass if we don't have values for them.

scripts/apiserver/resources.tmpl
dfeldick: The goal here was to only find code that had a var like this: Owner  interface{} within a struct and change it to this: Owner *apiv1.Owner.  This is exactly what is happening, so I'm not sure what your question is.

pkg/apic/apiserver/models/api/v1/resourceinstance.go
dfeldick: I think maybe. I've been doing some reading and testing. I'll do it.

jcollins-axway (approved):

sbolosan (approved):

scripts/apiserver/resources.tmpl
tsjohns9: I see, thanks for explaining. It was hard to read the if/else's on this line so I didn't really catch what was happening.

PR #200, opened 2021-08-23
APIGOV-54321 - api-server code generation improvements

jcollins-axway:
- Created a docker container for the base image that has all the tools necessary to generate the api server code
- Updated the scripts to expect the paths within the container rather than paths on disk
- Update make command to execute docker run rather than shell command
- Defaulted protocol, host, port to use production when not supplied
- Ran generation against prod for latest changes


kflan-io (approved):

vivekschauhan (approved):

dfeldick (approved):

dfeldick:
Looked at this a little more. 1), there is a file "openapitools.json in agent-sdk. With the move of the generator to docker, is this file still necessary (and it's entry in .gitignore)? Also, in the SDK is a README.md that outlines all of the prerequisites for the codegen and how to run it. This should be changed I guess?

jcollins-axway:
> Looked at this a little more. 1), there is a file "openapitools.json in agent-sdk. With the move of the generator to docker, is this file still necessary (and it's entry in .gitignore)? Also, in the SDK is a README.md that outlines all of the prerequisites for the codegen and how to run it. This should be changed I guess?

That file is set to be removed after each generation as it is created in the docker container while running the generator but not needed outside the container.  It just happens to be created in the volume we are mapping.

Good call on doc, I will look into updating that.  As that complicated process is no longer needed.

jcollins-axway:
> Looked at this a little more. 1), there is a file "openapitools.json in agent-sdk. With the move of the generator to docker, is this file still necessary (and it's entry in .gitignore)? Also, in the SDK is a README.md that outlines all of the prerequisites for the codegen and how to run it. This should be changed I guess?

I do not see any README with that documents that? Can you point me to it?

dfeldick:
https://github.com/Axway/agent-sdk/tree/main/pkg/apic/apiserver#readme

tsjohns9 (approved):

PR #206, opened 2021-08-31
[AMPGW-252] Model changes for latest amplify gateway deployment flow. Gov Agent changes in AMPGW-119.

mcollins-axway:
Also includes a commit for a change I needed to make to make the code more testable.


jcollins-axway (approved):

vivekschauhan (approved):

sbolosan (commented):
For me...the code looks good. Can we do a code walk through so I can get a higher level picture of how this works. Not sure if anyone else needs it too. Maybe just me 😳

PR #207, opened 2021-08-31
APIGOV-20279 - verify if email body is using {{.AuthTemplate}}

sbolosan:
Allow the customer to configure {{.AuthTemplate}} 
1.  this is an alternate solution for {{if}} / {{else}}
2.  this automatically pulls in the auth type. If it doesn't exist in the env_var, it will use the default.

### DIFF
--- a/pkg/config/subscriptionconfig.go
+++ b/pkg/config/subscriptionconfig.go
@@ -440,7 +440,7 @@ func (s *SubscriptionConfiguration) validateSubscriptionConfig() error {
 
 	for variable, template := range templates {
 		emailTemplate.IsAPIKey = !(variable == oauthEnvVar) // apikey for all but oauth
-		_, err := emailtemplate.ValidateSubscriptionConfig(template, "", emailTemplate, false)
+		_, err := emailtemplate.ValidateSubscriptionConfigOnStartup(template, "", emailTemplate)
 		if err != nil {
 			templateErr := fmt.Errorf("%s template is not valid: %s", variable, err.Error())
 			log.Error(templateErr)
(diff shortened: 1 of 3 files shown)


pkg/config/subscriptionconfig.go
```diff
@@ -440,7 +440,7 @@ func (s *SubscriptionConfiguration) validateSubscriptionConfig() error {
 
 	for variable, template := range templates {
 		emailTemplate.IsAPIKey = !(variable == oauthEnvVar) // apikey for all but oauth
-		_, err := emailtemplate.ValidateSubscriptionConfig(template, "", emailTemplate)
+		_, err := emailtemplate.ValidateSubscriptionConfig(template, "", emailTemplate, false)
```
dfeldick, line 443: I never like passing these true/false values to make a function do something different. Too easy for a coder to accidentally pick the wrong one, and if the code is to be self-documenting, you have no idea what the "false" on this line does. But, it works. The only alternative would be to to have a ValidateStartupConfig and ValidateRuntimeConfig (or whatever you name them), pass the original 3 params, then have these call private emailtemplate.validateSubscriptionConfig with t/f. It just hides it a little. Probably not worth the bother.

sbolosan: The calling func has a comment on what the bool means.  But I'll create separate, more explicit funcs

dfeldick: Up to you. I'm not gonna be a stickler about it.

dfeldick (approved):

pkg/config/subscriptionconfig.go
jcollins-axway: I like the separate function, because I think we will deprecate this in the next major (note that as we have) so clean up will be easier

jcollins-axway (approved):

vivekschauhan (approved):

PR #208, opened 2021-09-03
APIGOV-19801 - change some debug logs to trace

dfeldick:
Based on the ambiguity of debug vs trace and what customers want or need to see, this is about the extent I want to go for right now. Just move some of the messages that are clearly pretty low-level. If there are no objections and this is approved, I will take a similar approach to the agents. If you feel that it's gone overboard, let me know also.

### DIFF
--- a/docs/utilities/index.md
+++ b/docs/utilities/index.md
@@ -186,6 +186,8 @@ log.Infof("Log entry with format, %s", "additional log ~~message")
 
 log.Debugf("No changes detected in the API %s", *azAPI.Name)
 
+log.Trace("I got here in the code")
+
 log.Errorf("Error in processing : %s", err.Error())
 
 log.Warnf("Config not found: %s", missingItem)
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -191,7 +191,7 @@ func (c *agentRootCommand) initialize(cmd *cobra.Command, args []string) error {
 
 	viper.WatchConfig()
 	viper.OnConfigChange(func(e fsnotify.Event) {
-		log.Debugf("Config file changed : %s", e.Name)
+		log.Tracef("Config file changed : %s", e.Name)
 		c.onConfigChange()
 	})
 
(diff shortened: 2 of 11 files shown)


docs/utilities/index.md
```diff
@@ -186,6 +186,8 @@ log.Infof("Log entry with format, %s", "additional log ~~message")
 
 log.Debugf("No changes detected in the API %s", *azAPI.Name)
 
+log.Trace("I got here in the code")
+
```
sbolosan, line 190: what is this for?

pkg/cmd/root.go
```diff
@@ -191,7 +191,7 @@ func (c *agentRootCommand) initialize(cmd *cobra.Command, args []string) error {
 
 	viper.WatchConfig()
 	viper.OnConfigChange(func(e fsnotify.Event) {
-		log.Debugf("Config file changed : %s", e.Name)
+		log.Tracef("Config file changed : %s", e.Name)
```
sagarkal, line 194: I think this needs to be info, because its config change

docs/utilities/index.md
dfeldick: this is a file showing examples of how to use the log calls. There was no previous example of how to use Trace if a customer wanted to do such a thing in their agents.

sbolosan: ok...

sbolosan:
Good start...  i know i see some i created, so i'm guilty of abusing debug...

pkg/cmd/root.go
dfeldick: Cyrille, in another ticket, had a specific list of what should be info. This wasn't one. But maybe it should stay debug? I don't know.

sagarkal (approved):

pkg/cmd/root.go
sbolosan: I think Info or debug would work... since @sagarkal  is right, I think stuff like config changed should be seen by customer

jcollins-axway (approved):

pkg/cmd/root.go
dfeldick: put back to debug

PR #212, opened 2021-09-20
APIGOV-20509 - fix race conditions

dfeldick:
- moved IsNotTest from agent to util for more global availability
- called IsNotTest to prevent healthcheck startup during tests, which caused race conditions for tests
- lock poolStatus get/set, which were flagged as race conditions in tests

### DIFF
--- a/pkg/jobs/pool.go
+++ b/pkg/jobs/pool.go
@@ -20,6 +20,7 @@ type Pool struct {
 	jobsMapLock             sync.Mutex
 	cronJobsMapLock         sync.Mutex
 	detachedCronJobsMapLock sync.Mutex
+	poolStatusLock          sync.Mutex
 	failJobChan             chan string
 	stopJobsChan            chan bool
 	retryInterval           time.Duration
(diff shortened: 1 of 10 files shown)


pkg/jobs/pool.go
```diff
@@ -208,9 +209,18 @@ func (p *Pool) GetJobStatus(id string) string {
...
+	p.poolStatusLock.Lock()
+	defer p.poolStatusLock.Unlock()
 	return p.poolStatus.String()
 }
 
+//SetStatus - Sets the status of the pool of jobs
+func (p *Pool) SetStatus(status PoolStatus) {
+	p.poolStatusLock.Lock()
+	p.poolStatus = status
+	p.poolStatusLock.Unlock()
```
jcollins-axway, line 221: why not do the defer here like above?

dfeldick: Could. I thought that since it was only one line away it didn't matter. I can change if you wish.

jcollins-axway (approved):

pkg/jobs/pool.go
jcollins-axway: I just didn't like that its one way in Get and another in Set...your choice

dfeldick: Yeah, I did think about that. I'm changing it.

vivekschauhan (approved):

PR #215, opened 2021-09-23
APIGOV-20587 - warn that we could not get version from jfrog

sbolosan:
1. If we cannot connect to jfrog or decode htmlAnchors, return err
2. However, when we return error, warn only that we could not get version and continue instead of bubbling up error

### DIFF
--- a/pkg/cmd/agentversionjob.go
+++ b/pkg/cmd/agentversionjob.go
@@ -75,12 +75,15 @@ func (avj *AgentVersionCheckJob) Execute() error {
 	err = avj.getJFrogVersions(avj.urlName)
 	if err != nil {
 		log.Trace(err)
-		return err
+		// Could not get update from jfrog.  Warn that we could not determine version and continue processing
+		log.Warn("Agent cannot determine the next available release. Be aware that your agent could be outdated.")
+	} else {
+		// Successfully got jfrog version.  Now compare build to latest version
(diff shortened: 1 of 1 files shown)


pkg/cmd/agentversionjob.go
```diff
@@ -75,12 +75,15 @@ func (avj *AgentVersionCheckJob) Execute() error {
 	err = avj.getJFrogVersions(avj.urlName)
 	if err != nil {
 		log.Trace(err)
-		return err
-	}
-	// compare build to latest version
-	if isVersionStringOlder(avj.buildVersion, config.AgentLatestVersion) {
-		log.Warnf("New version available. Please consider upgrading from version %s to version %s", avj.buildVersion, config.AgentLatestVersion)
+		// Could not get update from jfrog.  Warn that we could not determine version and continue processing
+		log.Warn("Agent cannot determine the next available release. Be aware that your agent could be outdated.")
```
dfeldick, line 79: I like "out of date" rather than "outdated", but not a big deal either way.

dfeldick (approved):

pkg/cmd/agentversionjob.go
sbolosan: @dfeldick i'll check with C...  I mean, I say that milk could be outdated.  ;)

jcollins-axway (approved):

sagarkal (approved):

kflan-io (approved):

PR #219, opened 2021-09-29
APIGOV-20614 - create an traceability api (path) exceptions list

sbolosan:
1. new traceability configuration - apiExceptionsList: ${TRACEABILITY_EXCEPTION_LIST:"stringPath1, stringPath2, etc"}
2. Before redaction on URI, check to see if api path matches with entries in apiExceptionsList
3. If we find a match, do not send the event through
4. Should not affect go right path
5. Create debug logs to show what api path sdk found to ignore
6. Fix lint warning about upper case error verbiage(s)

### DIFF
--- a/pkg/transaction/definitions.go
+++ b/pkg/transaction/definitions.go
@@ -158,5 +158,5 @@ type Protocol struct {
 	ResponsePayload        string `json:"responsePayload,omitempty"`
 	WafStatus              int    `json:"wafStatus,omitempty"`
 	Timing                 string `json:"timing,omitempty"`
-	URIRaw                 string `json:"uriRaw,omitempty"`
+	uriRaw                 string
 }
--- a/pkg/transaction/definitions.go
+++ b/pkg/transaction/definitions.go
@@ -158,4 +158,5 @@ type Protocol struct {
 	ResponsePayload        string `json:"responsePayload,omitempty"`
 	WafStatus              int    `json:"wafStatus,omitempty"`
 	Timing                 string `json:"timing,omitempty"`
+	URIRaw                 string `json:"uriRaw,omitempty"`
 }
--- a/pkg/transaction/eventgenerator.go
+++ b/pkg/transaction/eventgenerator.go
@@ -119,7 +119,8 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
 	events := make([]beat.Event, 0)
 
 	// See if the uri is in the api exceptions list
-	if len(detailEvents) > 0 && e.isInAPIExceptionsList(detailEvents) {
+	if e.isInAPIExceptionsList(detailEvents) {
+		log.Debug("Found api path in traceability api exceptions list.  Ignore transaction event")
 		return events, nil
 	}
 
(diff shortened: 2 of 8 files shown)


pkg/transaction/eventgenerator.go
```diff
@@ -117,6 +118,11 @@ func (e *Generator) createEvent(logEvent LogEvent, eventTime time.Time, metaData
 func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent, eventTime time.Time, metaData common.MapStr, eventFields common.MapStr, privateData interface{}) ([]beat.Event, error) {
 	events := make([]beat.Event, 0)
 
+	// See if the uri is in the api exceptions list
+	if len(detailEvents) > 0 && e.isInAPIExceptionsList(detailEvents) {
```
dfeldick, line 122: I would put the len check in the isInAPI function. This way anyone who ever might try to call the func doesn't have to always check the len also. The func should not nil ptr out if it gets passed an empty array.

sbolosan: i had it like that originally... i guess i can put it back.
No one should be really calling this except the func its in now.  But I get it.

pkg/transaction/eventgenerator.go
```diff
@@ -170,6 +178,34 @@ func (e *Generator) createSamplingTransactionDetails(summaryEvent LogEvent) samp
...
+	apiExceptionsList := traceability.GetAPIExceptionsList()
+	exceptions := strings.Split(apiExceptionsList, ",")
+	for i := range exceptions {
+		exceptions[i] = strings.TrimSpace(exceptions[i])
+	}
+
+	// If the api path exists in the exceptions list, return true and ignore event
+	for _, value := range exceptions {
+		if value == uriRaw {
+			log.Debugf("Found api path %s in traceability api exceptions list.  Ignore transaction event", uriRaw)
```
dfeldick, line 201: Maybe the caller of this func should do the logging. The func returns T/F based on whether or not something is in the list. It could be called somewhere else for a completely different purpose, so this func shouldn't be in charge of deciding and logging that a transaction event is going to be ignored.

sbolosan: i don't understand what you're saying?  what other purpose could it be called for, except checking if the path is in the list?

dfeldick: The function is answering the question "is it in the list, yes or no". The first part of the debug is OK. But the second part about ignoring the event is a decision the caller of the function is making, so IMO that is where the "I'm gonna ignore this" should be. I know it's a little picky, but what if someone called the function only for the purpose of checking to see if something is in the list (which is what the function name implies). If it's in the list they want to do something, and if not something else. But in neither case are they going to "ignore transaction event". They will still see that in the logs.

sbolosan: i lose the path to display since i'm only returning bool...  i guess i can return both?  i'll figure it out.

sbolosan:
@jcollins-axway @vivekschauhan when you guys get a chance to make sure I didn't break anything 😬

vivekschauhan (approved):

pkg/transaction/definitions.go
```diff
@@ -158,4 +158,5 @@ type Protocol struct {
 	ResponsePayload        string `json:"responsePayload,omitempty"`
 	WafStatus              int    `json:"wafStatus,omitempty"`
 	Timing                 string `json:"timing,omitempty"`
+	URIRaw                 string `json:"uriRaw,omitempty"`
```
jcollins-axway, line 161: URI raw should not be marshaled to json.  Also, does it need to be public?

jcollins-axway (changes requested):
remove the json definition of uriraw in protocol struct

dfeldick (approved):

jcollins-axway (approved):

PR #224, opened 2021-10-08
APIGOV-20511 - publish previous state when agent state changes

jcollins-axway:


jcollins-axway:
@sbolosan @dfeldick @vivekschauhan @Atraya21 @tsjohns9 @sagarkal - if you all have time

PR #240, opened 2021-11-10
APIGOV-20960 - use actual constant prefix

sbolosan:


dfeldick (approved):

sbolosan:
@vivekschauhan , i was missing the actual secret check... I already committed and merged.  Can you just take a quick peek

PR #245, opened 2021-11-15
APIGOV-20957 Cache events from a grpc stream

tsjohns9:


sbolosan (approved):

kflan-io (approved):

pkg/agent/stream/handlers_test.go
```diff
@@ -0,0 +1,400 @@
...
+				assert.NotNil(t, err)
+			} else {
+				assert.Nil(t, err)
+			}
+		})
+	}
+
+}
+
+type mockCache struct {
```
jcollins-axway, line 312: Just a thought, maybe we move the mock cache struct into the cache package, making it public, so other tests can use it.

pkg/agent/agent.go
```diff
@@ -189,20 +191,37 @@ func startAPIServiceCache() {
...
+		stream.NewCategoryHandler(agent.categoryMap),
+		// TODO: agents should be able to pass in their own handlers.
+	)
+
+	go func() {
+		err := c.Start()
+		if err != nil {
+			log.Errorf("failed to start the grpc client: %s", err)
+		}
+	}()
```
jcollins-axway, line 224: Can we pull this out into a separate function like the discovery cache function?

pkg/apic/auth/apicauth.go
```diff
@@ -529,10 +529,10 @@ func (ctg *channelTokenGetter) Close() error {
 	return nil
 }
 
-// TokenAuth implements the PerRPCCredentials interface
+// TokenAuth -
 type TokenAuth struct {
-	tenantID       string
-	tokenRequester PlatformTokenGetter
+	TenantID       string
+	TokenRequester TokenGetter
```
jcollins-axway, line 535: I may have missed it, but did not see why this is not public

jcollins-axway (approved):

pkg/apic/auth/apicauth.go
tsjohns9: I left these public because the `NewTokenAuth` method sets the `TokenRequester` field and I want to be able to use my own. In this case It's the `TokenGetter` on the agent.

jcollins-axway: Makes sense, thank you sir

pkg/agent/agent.go
tsjohns9: Yes

jcollins-axway (approved):

jcollins-axway (approved):

sbolosan (approved):

PR #247, opened 2021-11-16
APIGOV-20991 - leave token locking and debug updates

sbolosan:
Based on iteration triage with Peter O.  deciding to leave token locking in

### DIFF
--- a/pkg/apic/auth/apicauth.go
+++ b/pkg/apic/auth/apicauth.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 	"time"
 
 	"crypto/rsa"
@@ -97,6 +98,7 @@ func NewPlatformTokenGetter(privKey, publicKey, password, url, aud, clientID str
 			password:  password,
 		},
 		&tokenHolder{},
+		&sync.Mutex{},
 	}
 }
 
@@ -117,6 +119,7 @@ func NewPlatformTokenGetterWithCentralConfig(centralCfg config.CentralConfig) Pl
 			password:  centralCfg.GetAuthConfig().GetKeyPassword(),
 		},
 		&tokenHolder{},
+		&sync.Mutex{},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/apic/auth/apicauth.go
```diff
@@ -333,10 +336,9 @@ func (ptg *platformTokenGenerator) getPlatformTokens(requestToken string) (*axwa
 
 	defer closeHelper(resp.Body)
 	if resp.StatusCode != 200 {
-		body, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			log.Debugf("bad response from AxwayID: %s: %s", resp.Status, body)
-		}
+		body, _ := ioutil.ReadAll(resp.Body)
+		log.Debugf("bad response from AxwayID: %s: %s, request token :%s, request time : %s", resp.Status, body, requestToken, startTime.String())
```
jcollins-axway, line 340: Do we want to log the token here, is that not a security issue?

sbolosan: oops... you're right.  this is left over from Peter's binary... thanks!

pkg/apic/auth/apicauth.go
```diff
@@ -374,6 +376,7 @@ type platformTokenGetter struct {
 	*platformTokenGenerator
 	*keyReader
 	*tokenHolder
+	getTokenMutex *sync.Mutex
```
jcollins-axway, line 379: My only issue with the mutex being here is if multiple processes create different token getters.  All we are locking is that the one instantiation is not requesting multiple at a time.  If the lock was a global in the package we can ensure that only one token getter can request a token at a time. Not sure if that matters or not, however.  @vivekschauhan thoughts?

vivekschauhan: I think, locking on the instance should be good. Globally locking will lock each instance which might not be expected as instances can be configured with different configuration.

vivekschauhan (approved):

PR #266, opened 2021-12-09
APIGOV-21043 - service builder changes for Apigee agent

jcollins-axway:
- added new service body attribute for changing the descriptor of a stage on an API Service and Catalog, default Stage
- added ability to add different attributes to API Service vs API Revision vs API Instance


sbolosan (approved):
Like the maps...

sbolosan (approved):

vivekschauhan (approved):

sbolosan (approved):

dfeldick (approved):

PR #268, opened 2021-12-15
[AMPGW-569] Add agentFeatures config section to allow disabling of standard agent features

mcollins-axway:
…DK for services that are not connected to Platform.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -347,7 +347,7 @@ func getAgentResource() (*apiV1.ResourceInstance, error) {
 
 // updateAgentStatus - Updates the agent status in agent resource
 func updateAgentStatus(status, prevStatus, message string) error {
-	if !agent.cfg.GetConnected() {
+	if agent.cfg == nil || !agent.cfg.GetConnected() {
 		return nil
 	}
 	// IMP - To be removed once the model is in production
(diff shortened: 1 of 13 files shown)


sbolosan (approved):

sbolosan (approved):

pkg/agent/agent.go
```diff
@@ -74,6 +75,11 @@ var agent = agentData{}
 
 // Initialize - Initializes the agent
 func Initialize(centralCfg config.CentralConfig) error {
+	return InitializeWithAgentFeatures(centralCfg, config.NewAgentFeaturesConfiguration())
```
vivekschauhan, line 78: Thanks for keeping this.

pkg/config/agentfeaturesconfig.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"github.com/Axway/agent-sdk/pkg/cmd/properties"
+)
+
+// AgentFeaturesConfig - Interface to get agent features Config
+type AgentFeaturesConfig interface {
+	ConnectionToCentralEnabled() bool
+	ProcessSystemSignalsEnabled() bool
+	AgentVersionCheckerEnabled() bool
```
vivekschauhan, line 11: May be we can remove the version checker config in CentralConfig

vivekschauhan (approved):

jcollins-axway (approved):

pkg/config/agentfeaturesconfig.go
mcollins-axway: Agreed it should go. Was afraid I'd break something, but suspect the Amplify Gateway is the only user of this flag. Will I just remove it ?

PR #271, opened 2021-12-20
APIGOV-21297 - Watch for Agent resource

vivekschauhan:
- Updates to watch for update events on agent resource
- Includes refactoring to manage agent resource


pkg/agent/handler/handlers.go
```diff
@@ -107,3 +117,29 @@ func (c categoryHandler) handle(action proto.Event_Type, resource *apiv1.Resourc
...
+func (h *agentResourceHandler) Handle(action proto.Event_Type, resource *v1.ResourceInstance) error {
+	if h.agentResourceManager != nil && action == proto.Event_UPDATED {
+		kind := resource.Kind
+		switch kind {
+		case discoveryAgent:
+			fallthrough
+		case traceabilityAgent:
+			fallthrough
+		case governanceAgent:
+			h.agentResourceManager.SetAgentResource(resource)
```
jcollins-axway, line 141: Do other resource kinds come in here?  Is this switch necessary?

jcollins-axway (approved):

pkg/agent/handler/handlers.go
vivekschauhan: The handler gets invoked for all resource events received by the watch client, so this handler does the check for agent resource kind to process them further

pkg/agent/resource/manager.go
```diff
@@ -0,0 +1,232 @@
...
+	case v1alpha1.TraceabilityAgentResourceName:
+		agentRes := createTraceabilityAgentStatusResource(a.cfg.GetAgentName(), status, prevStatus, message)
+		resourceInstance, _ := agentRes.AsInstance()
+		return resourceInstance
+	case v1alpha1.GovernanceAgentResourceName:
+		agentRes := createGovernanceAgentStatusResource(a.cfg.GetAgentName(), status, prevStatus, message)
+		resourceInstance, _ := agentRes.AsInstance()
+		return resourceInstance
+	default:
+		panic(ErrUnsupportedAgentType)
```
tsjohns9, line 212: Can we return an error instead of a panic?

tsjohns9 (approved):

pkg/agent/resource/manager.go
vivekschauhan: Done

PR #274, opened 2021-12-21
APIGOV-19455 Add a field in transaction summary event to indicate tha…

sagarkal:
…t the transaction is already covered by usage metrics

### DIFF
--- a/pkg/transaction/definitions.go
+++ b/pkg/transaction/definitions.go
@@ -62,16 +62,16 @@ type LogEvent struct {
 
 // Summary - Represent the transaction summary event
 type Summary struct {
-	Status                       string       `json:"status,omitempty"`
-	StatusDetail                 string       `json:"statusDetail,omitempty"`
-	Duration                     int          `json:"duration"`
-	Application                  *Application `json:"application,omitempty"`
-	Product                      *Product     `json:"product,omitempty"`
-	Team                         *Team        `json:"team,omitempty"`
-	Proxy                        *Proxy       `json:"proxy,omitempty"`
-	Runtime                      *Runtime     `json:"runtime,omitempty"`
-	EntryPoint                   *EntryPoint  `json:"entryPoint,omitempty"`
-	IsIncludedInUsageMetricEvent bool         `json:"is_included_in_usage_metric_event"`
+	Status          string       `json:"status,omitempty"`
+	StatusDetail    string       `json:"statusDetail,omitempty"`
+	Duration        int          `json:"duration"`
+	Application     *Application `json:"application,omitempty"`
+	Product         *Product     `json:"product,omitempty"`
+	Team            *Team        `json:"team,omitempty"`
+	Proxy           *Proxy       `json:"proxy,omitempty"`
+	Runtime         *Runtime     `json:"runtime,omitempty"`
+	EntryPoint      *EntryPoint  `json:"entryPoint,omitempty"`
+	IsInMetricEvent bool         `json:"isInMetricEvent,omitempty"`
 }
 
 // Application  - Represents the application used in transaction summary event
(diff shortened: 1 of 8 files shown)


pkg/transaction/definitions.go
```diff
@@ -62,16 +62,16 @@ type LogEvent struct {
...
+	Status                       string       `json:"status,omitempty"`
+	StatusDetail                 string       `json:"statusDetail,omitempty"`
+	Duration                     int          `json:"duration"`
+	Application                  *Application `json:"application,omitempty"`
+	Product                      *Product     `json:"product,omitempty"`
+	Team                         *Team        `json:"team,omitempty"`
+	Proxy                        *Proxy       `json:"proxy,omitempty"`
+	Runtime                      *Runtime     `json:"runtime,omitempty"`
+	EntryPoint                   *EntryPoint  `json:"entryPoint,omitempty"`
+	IsIncludedInUsageMetricEvent bool         `json:"is_included_in_usage_metric_event"`
```
vivekschauhan, line 74: Can we have a short name for the attribute?

vivekschauhan (approved):

pkg/transaction/definitions.go
sagarkal: Spent quite some time on this variable name 😄 how about just _IsIncludedInUsageMetric_?

kflan-io: isInUsageMetric?

kflan-io (approved):

pkg/transaction/definitions.go
vivekschauhan: Not concerned about the variable name but the attribute name "is_included_in_usage_metric_event". Can we use "hasMetricEvent" or something similar instead?

pkg/transaction/logeventbuilder.go
```diff
@@ -84,7 +84,7 @@ type SummaryBuilder interface {
 	SetProxy(proxyID, proxyName string, proxyRevision int) SummaryBuilder
 	SetRunTime(runtimeID, runtimeName string) SummaryBuilder
 	SetEntryPoint(entryPointType, method, path, host string) SummaryBuilder
-
+	SetIsIncludedInUsageMetricEvent(isIncludedInUsageMetricEvent bool) SummaryBuilder
```
jcollins-axway, line 87: Is this builder method still necessary if we are only concerned with the attribute being set when the new summary builder is instantiated?

pkg/transaction/definitions.go
sagarkal: My idea was that if someone looks at a transaction summary event they can immediately figure out what this attribute exactly means. What if we use _IsInUsageMetric_ as variable name and _is_in_usage_metric_ as attribute name?

pkg/transaction/logeventbuilder.go
sagarkal: I thought its a standard to have the builder method for all the attributes, because I saw that there is a builder method for fields like APICDeployment, EnvironmentID etc as well that are only set when the new summary builder is instantiated

jcollins-axway: That makes sense, then we can keep it and eventually clean up the extra setters.

jcollins-axway (approved):

pkg/transaction/definitions.go
sagarkal: Changed it to _IsInUsageMetric_ & _is_in_usage_metric_ respectively

vivekschauhan: Can we make the attribute name as well in camel case (isInUsageMetric instead of is_in_usage_metric) to be consistent?

jcollins-axway: I think the final change is already in there as isInUsageMetric for both names

jcollins-axway (approved):

PR #278, opened 2022-01-06
APIGOV-21351 - Support for registering event handlers

vivekschauhan:
- Added methods to allow agent implementation to register/unregister event handlers
- Added Proxy resource handler to invoke registered event handlers. Proxy handler gets registered to event listener during the agent initialization, this allows agent implementation to register event handlers when needed(before agent initialization or while agent is running)
- Updates to pass event metadata to handlers

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -61,10 +61,13 @@ type agentData struct {
 	deleteServiceValidator     DeleteServiceValidator
 	configChangeHandler        ConfigChangeHandler
 	agentResourceChangeHandler ConfigChangeHandler
+	proxyResourceHandler       handler.ProxyHandler
 	isInitialized              bool
 }
 
-var agent = agentData{}
+var agent = agentData{
+	proxyResourceHandler: handler.NewProxyHandler(),
+}
 
 // Initialize - Initializes the agent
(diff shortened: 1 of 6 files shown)


tsjohns9 (approved):

pkg/agent/agent.go
```diff
@@ -344,14 +357,20 @@ func startDiscoveryCache(instanceCacheLock *sync.Mutex) {
 }
 
 func startStreamMode(agent agentData) error {
-	cs, err := stream.NewStreamer(
-		api.NewClient(agent.cfg.GetTLSConfig(), agent.cfg.GetProxyURL()),
-		agent.cfg,
-		agent.tokenRequester,
+	proxyHandler := agent.proxyResourceHandler.(handler.Handler)
```
tsjohns9, line 360: Not that big a deal, but instead of doing a type conversion we could have `NewProxyHandler` return the struct which implements both interfaces, and `agent.proxyResourceHandler` would be set to the struct instead of the `ProxyHandler` Interface.

vivekschauhan: Thanks Trevor

tsjohns9 (approved):

PR #283, opened 2022-01-13
APIGOV-20073 - Support for gRPC based watch for API server resources

vivekschauhan:
* APIGOV-20123 - Watch service RPC definition

* APIGOV-20701 - Changes to use WatchTopic resource for registering the subscription with watch service

* APIGOV-20812 - Added watch manager to register and subscribe to watch events
- Implementation can create an instance of watch manager and register multiple subscriptions by calling RegisterWatch with watch topic
- The RegisterWatch takes two channels as argument for event and to raise errors
- Added an example client under samplesEnable watch-controller

* APIGOV-20825 - Match inbound events based on the watch topic self link

* APIGOV-20992 - Update watch service package, service and method name

* APIGOV-20957 - Updates to use watch events to update agent caches

* APIGOV-21020 - Changes to include resource attribute in watch events

* APIGOV-20956 - Option to switch between grpc or poll mode

* APIGOV-20958 - Update Central healthcheck when running in grpc mode

* APIGOV-21015 - Updates to use consistent apiserviceinstance validation in poll and grpc mode

* APIGOV-21060 - Reconnect the watch client on healthcheck failure

* APIGOV-21059 - Create watch topic at agent initialization

* APIGOV-21058 - Integration with harvester to sync events
- Added a client for harvester service to fetch events based on the sequence ID
- Updated watch manager to fetch events based on the cached sequence ID before starting to receive events on gRPC stream

* APIGOV-21297 - Updates to watch for update events on agent resource
- Includes refactoring to manage agent resource

* APIGOV-21303 - Restart the stream to the watch controller when the connection closes

* APIGOV-21298 - Proxy support for gRPC watch and harvester connections

APIGOV-21351 - Support for registering event handlers
- Added methods to allow agent implementation to register/unregister event handlers
- Added Proxy resource handler to invoke registered event handlers. Proxy handler gets registered to event listener during the agent initialization, this allows agent implementation to register event handlers when needed(before agent initialization or while agent is running)

* APIGOV-21057 -  Updates for cache management
- Added agent cache manager to use caches consistently in poll or stream mode
- Updated event listener to cache the sequence ID of the last processed event for the watch topic to allow subsequent startups to read events from harvester
- Added config for agent cache storage path and interval to trigger saving the cache
- Added config to specify host and port of GRPC service. No defaults on the config overrides using the host and port from the central url config.


pkg/agent/agent.go
```diff
@@ -41,38 +45,27 @@ type APIValidator func(apiID, stageName string) bool
...
+	cacheManager               agentcache.Manager
 	apiValidator               APIValidator
 	configChangeHandler        ConfigChangeHandler
 	agentResourceChangeHandler ConfigChangeHandler
+	proxyResourceHandler       *handler.StreamWatchProxyHandler
 	isInitialized              bool
 }
 
-var agent = agentData{}
+var agent = agentData{
```
jcollins-axway, line 66: Should the agent var here be initialized in an init func so its only created on the first import of the agent package? Does that matter?

pkg/agent/agent.go
```diff
@@ -111,41 +97,49 @@ func InitializeWithAgentFeatures(centralCfg config.CentralConfig, agentFeaturesC
...
-		// Init apic client
-		if agent.apicClient == nil {
-			agent.apicClient = apic.New(centralCfg, agent.tokenRequester)
-			agent.apicClient.AddCache(agent.categoryMap, agent.teamMap)
-		} else {
-			agent.apicClient.SetTokenGetter(agent.tokenRequester)
-			agent.apicClient.OnConfigChange(centralCfg)
+		// Init apic client when the agent starts, and on config change.
+		agent.apicClient = apic.New(centralCfg, agent.tokenRequester)
+		agent.apicClient.AddCache(agent.cacheManager.GetCategoryCache(), agent.teamMap)
```
jcollins-axway, line 121: If we add the teamMap to the cache manager then maybe we just send in a ref to the cache manager with the New apic client?

jcollins-axway: Or, maybe better, make adding these caches using the functional options pattern?

pkg/apic/apiserver/models/management/v1alpha1/WatchTopic.go
```diff
@@ -0,0 +1,99 @@
...
+		GroupKind: apiv1.GroupKind{
+			Group: "management",
+			Kind:  "WatchTopic",
+		},
+		APIVersion: "v1alpha1",
+	}
+)
+
+const (
+	WatchTopicScope = ""
```
jcollins-axway, line 24: The generator was updated to make this an array of scopes, I do not think it will matter for this resource, however.

pkg/apic/client.go
```diff
@@ -335,7 +349,7 @@ func (c *ServiceClient) checkAPIServerHealth() error {
 
 	if c.cfg.GetTeamID() == "" {
 		// Validate if team exists
-		team, err := c.getCentralTeam(c.cfg.GetTeamName())
+		team, err := c.GetCentralTeamByName(c.cfg.GetTeamName())
```
jcollins-axway, line 352: Is this needed anymore with the addition in the agent package that was made? initEnvResources

sbolosan (approved):

pkg/watchmanager/config.go
```diff
@@ -0,0 +1,32 @@
+package watchmanager
```
jcollins-axway, line 1: We have a config package, should this be there?

jcollins-axway (approved):

pkg/agent/agent.go
vivekschauhan: We can have a separate story to move the team map to cache manger. Once we move the team cache to cache manager, we can remove the AddCache method from service client. The reason I think we added them there was to avoid cyclic references and with agent cache management moved to different package that can be avoided.

pkg/apic/apiserver/models/management/v1alpha1/WatchTopic.go
vivekschauhan: We will re-generate the source, but you are right it should not matter for WatchTopic resource

pkg/watchmanager/config.go
vivekschauhan: Wanted to keep the config related to watch manager completely separate and independent of other SDK package(some still used for harvester client though) to keep this as independent component.

pkg/apic/client.go
vivekschauhan: You are right, the checkAPIServerHealth method is left over, will remove the method as it is not used anymore

pkg/agent/agent.go
vivekschauhan: Done

pkg/apic/client.go
vivekschauhan: Done

Makefile
```diff
@@ -46,3 +46,14 @@ apiserver-generate: # generate api server resources, prod by default. ex: make a
...
 unifiedcatalog-generate: ## generate unified catalog resources
 	./scripts/unifiedcatalog/unifiedcatalog_generate.sh
+
+
+PROTOFILES := $(shell find $(WORKSPACE)/proto -type f -name '*.proto')
+PROTOTARGETS := $(PROTOFILES:.proto=.pb.go)
+
+%.pb.go : %.proto
+	@echo $<
+	@protoc  --proto_path=$(WORKSPACE)/proto --go-grpc_out=. --go_out=. $<
```
tsjohns9, line 56: I think we need to update this to properly output the files to the directory they are in now. I was doing it manually every time.

vivekschauhan: Done

tsjohns9 (approved):

jcollins-axway:
@tsjohns9 had an issue yesterday with the ACL job, it was failing because the central health check was no longer created.  Has there been testing in poll only mode?

PR #291, opened 2022-01-24
APIGOV-21460 - Updates for healthcheck based on gRPC connection

vivekschauhan:
- In gRPC mode, register central healthcheck based on the gRPC connection
- Updated jobs library to allow registering detachable channel job.
- Includes changes to register streamer client job as detachable channel job to allow recovering from gRPC connection failures. The job gets registered on initialization and on gRPC connection failure attempts to restart the streamer using a backoff interval

### DIFF
--- a/pkg/agent/stream/client.go
+++ b/pkg/agent/stream/client.go
@@ -5,9 +5,11 @@ import (
 	"net"
 	"net/url"
 	"strconv"
+	"time"
 
 	"github.com/Axway/agent-sdk/pkg/api"
 	"github.com/Axway/agent-sdk/pkg/apic/auth"
+	"github.com/Axway/agent-sdk/pkg/util"
 	"github.com/Axway/agent-sdk/pkg/util/log"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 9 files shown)


sbolosan (approved):

jcollins-axway (approved):

pkg/agent/stream/client.go
```diff
@@ -180,6 +218,7 @@ func (c *streamer) Start() error {
 	if err != nil {
 		return err
 	}
+	hc.RegisterHealthcheck(util.AmplifyCentral, "central", c.healthcheck)
```
tsjohns9, line 221: Can the RegisterHealthCheck func be added as a field on *streamer so that it is easier to test without making an actual call to that package in a test?

pkg/jobs/channeljob.go
```diff
@@ -15,6 +15,20 @@ type channelJob struct {
 	channelJobProps
 }
 
+//newDetachedChannelJob - creates an channel job, detached from other cron jobs
```
tsjohns9, line 18: creates **a** channel job

tsjohns9 (approved):

pkg/agent/stream/client.go
vivekschauhan: Moved the healthcheck registration as a callback which gets invoked after stream connection. We cannot register the healthcheck while creating the stream job since the registration will kick off a initial heathcheck making the central healthcheck failed and since the jobs relying on it would start to report not ready, the global pool will go in stopped state on agent initialization. So we need the initial registration done at the time of first stream connection.

pkg/jobs/channeljob.go
vivekschauhan: Done

PR #293, opened 2022-01-25
APIGOV-21478 Remove UpdateFromAPIServer cfg value

tsjohns9:
- Remove the UpdateFromAPIServer config value
- Exported functions to get API's by attributes in discovery.go will only check the sdk cache


tsjohns9:
Is there documentation that needs to be changed since I removed a config value?

sbolosan (approved):
Looks like there was docs associated with the original ticket... let me look

sbolosan:
@tsjohns9 , maybe not...  I only see api server reference to version and pattern stuff I did... @jcollins-axway did the original PT-SMI ticket, maybe he can shed some light on doc stuff, but as I see it, I don't see anything for pathUpdateFromAPIServer

jcollins-axway:
> @tsjohns9 , maybe not... I only see api server reference to version and pattern stuff I did... @jcollins-axway did the original PT-SMI ticket, maybe he can shed some light on doc stuff, but as I see it, I don't see anything for pathUpdateFromAPIServer

This was never a published option, so no docs to update

jcollins-axway (approved):

PR #300, opened 2022-01-28
APIGOV-21493_tmp create definitions package

tsjohns9:
My original APIGOV-21493 branch was branched off of grpc-dev. Trying to pull in main has many merge conflicts. I'm looking to merge this change to main to try and make my merge conflicts easier. For my branch I needed to move a few types to a new package to avoid a circular dependency. Moving the types caused changes in a lot of files, so if I can get this into main and then try to merge again it should be easier.

### DIFF
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -8,9 +8,9 @@ import (
 	"strings"
 	"testing"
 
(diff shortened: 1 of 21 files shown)


sbolosan (approved):
looks like a valid replace...

PR #302, opened 2022-01-31
APIGOV-21403 - add ability to get owner info from apimap cache

jcollins-axway:


sbolosan:
👍

sbolosan (approved):

pkg/agent/cache/manager.go
```diff
@@ -408,21 +408,21 @@ func (c *cacheManager) GetTeamByName(name string) *definitions.PlatformTeam {
 	if err != nil {
 		return nil
 	}
-	team, ok := item.(*definitions.PlatformTeam)
+	team, ok := item.(definitions.PlatformTeam)
```
tsjohns9, line 411: in the `AddTeam` func, you are saving a *Team. How come you dropped the type cast to the pointer?

tsjohns9: Nevermind, I see a *Team is passed in, and then you save the value to the cache, not the pointer. Can you save the pointer to the cache instead? It looks like that is how we are saving the other values.

jcollins-axway: The issue I just hit with saving a pointer to the cache is that the cache item at that point is mutable.  

In this case the teamcache job was looping over teams, assigning each to the same loop variable team.  That team was sent in to AddTeam as a pointer and was saved to the cache as such. Subsequent teams in that loop used the same variable, address, and were changing the team at that address., so the last team in the loop was the team at that address.

Then retrieving a team from the cache would result in the same address being given and then getting the same team object.

Perhaps if other places are setting the cache items as pointers we need to update those so the same issue can not occur.

tsjohns9: That makes sense

vivekschauhan: Can we not use the address of the array element(platformTeams) instead of the local variable in the call to AddTeam

for i, team := range platformTeams {
		savedTeam := j.cache.GetTeamByID(team.ID)
		if savedTeam == nil {
			j.cache.AddTeam(&platformTeams[i])
...
}

vivekschauhan (approved):

pkg/agent/cache/manager.go
jcollins-axway: @vivekschauhan we can make this change instead. Perhaps we should look into updating the cache to store the value of pointers, rather than the pointer themselves, so the cache is immutable. Would you agree to making a change like that?

vivekschauhan: Sure, let keep it the way you got it right now

PR #305, opened 2022-02-01
APIGOV-21514 - check api exceptions list

sbolosan:
Update api exceptions list for ta
1. allow api exceptions to contain wildcards (regular expression)
2. validate exception.  If validation fails, return error
3. when generating events, check to see if api exception matches url.  If it does, ignore generation


dfeldick (approved):

jcollins-axway (changes requested):

tsjohns9 (approved):

kflan-io (approved):

pkg/transaction/eventgenerator.go
```diff
@@ -221,7 +222,8 @@ func (e *Generator) isInAPIExceptionsList(logEvents []LogEvent) bool {
 
 	// If the api path exists in the exceptions list, return true and ignore event
 	for _, value := range exceptions {
-		if value == uriRaw {
+		match, _ := regexp.MatchString(value, uriRaw)
```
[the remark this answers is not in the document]
sbolosan: updated changes

sbolosan:
@jcollins-axway @vivekschauhan , updated and moved method into traceability package.  I guess I could have created an apiExceptionList.go, but decided to just leave the compile and regex list in config.  If you guys think this warrants moving this out into an entirely new go file, I can do that (later).

jcollins-axway:
> @jcollins-axway @vivekschauhan , updated and moved method into traceability package. I guess I could have created an apiExceptionList.go, but decided to just leave the compile and regex list in config. If you guys think this warrants moving this out into an entirely new go file, I can do that (later).

Let's get Vivek's reply, with that said...I would like to see this moved, imho.

sbolosan:
@jcollins-axway @vivekschauhan , i went ahead and just moved the content to apiExceptionList.go.  Let me know if you think there's any more refactoring to be done.  I need to test it, but somehow, my debug isn't being called.  Not sure why.  I'll jump back on that manana.

pkg/traceability/apiExceptionList.go
```diff
@@ -0,0 +1,58 @@
...
+// getAPIExceptionsList - Returns traceability APIs exception list (api paths)
+func getAPIExceptionsList() []*regexp.Regexp {
+	if newShowRegex == nil {
+		return []*regexp.Regexp{}
+	}
+	return newShowRegex
+}
+
+// newShowRegex - array of regexp.Regexp
+var newShowRegex []*regexp.Regexp
```
vivekschauhan, line 42: Can we wrap this variable in an struct/interface for managing the exception list and then setup can construct it? Something like
type APIExceptionList interface {
    ShouldIgnoreEvent(uriRaw string) bool // or some better method name 
}

type apiExceptionList struct {
    exceptionRegEx []*regexp.Regexp
}

func (al *apiExceptionList) ShouldIgnoreEvent(uriRaw string) bool {
  // use al.exceptionReg list to perform check
   ...
}

Then setupAPIExceptionList() can construct the apiExceptionList instance based on the config.

Don't like keeping global variable, but I understand for this case we will need to keep global variable for APIExceptionList interface as the object is constructed while parsing the traceability config.

vivekschauhan: Nevermind, we will anyway have to keep a global variable and then will have to add a method to return the constructed interface. Would have been good if we could store the constructed interface somewhere that can be referenced by other packages.

So, please ignore my comment

sbolosan: @vivekschauhan no worries.. Still updated to pear down unneeded code.  Re-check when you can... thank you.

PR #306, opened 2022-02-02
APIGOV-21518 Add marshal and unmarshal funcs to api server resources

tsjohns9:
- Allow all api server resources to set their own custom sub resources
- Add SubResource map to ResourceMeta struct
- Add a marshal and unmarshal func to resources.tmpl for generating the models
- Add a marshal and unmarshal func to ResourceMeta
- Update ResourceMeta marshal and unmarshal to handle sub resources


vivekschauhan (approved):

scripts/apiserver/resources.tmpl
```diff
@@ -99,4 +77,87 @@ func (res *{{.res.kind}}) AsInstance() (*apiv1.ResourceInstance, error) {
...
+        return err
+    }
+
+    {{range $field, $isNotEmpty := .res.fields -}}
+        {{- $fieldName := $field | strings.Title -}}
+        {{- $keyName := $field | strings.ToLower -}}
+        {{if eq $keyName "owner"}}
+        {{else if eq $keyName "spec"}}
+        {{else}}
+        sr, err = json.Marshal(aux.SubResources["{{$keyName}}"])
```
jcollins-axway: could we throw a comment here to note what sub resource is being handled?

// Unmarshalling subresource {{$fieldName}}

....or similar?

Maybe one above the spec handling as well?

jcollins-axway (approved):

scripts/apiserver/resources.tmpl
tsjohns9: Good idea

jcollins-axway: isn't the action unmarshaling?

jcollins-axway: that step is marshal, but overall we want to unmarshal the sub resource

jcollins-axway: I guess it is not necessary, but should we regen?

PR #309, opened 2022-02-03
[STREAMS-2440] - add support of integer,number,object,array subscription property builder

epiasentin-axway:
We are working on Streams Discovery Agent and in order to manage Streams subscriptions in the Unified Catalog, we have to add the support of complex subscription model like this:
```
{
    "subscriptionMode": "event",
    "webhookUrl": "<YOUR_WEBHOOK_ENDPOINT_URL>",
    "subscriptionStatus": "active",
    "retryOnHttpCodes": [500,503,504],
    "retryMaxAttempts": 10,
    "retryBackOffInitialDuration": "PT5S",
    "retryBackOffMaxDuration": "PT30S",
    "retryBackOffFactor": 1,
    "webhookHeaders": {
        "X-Custom-Header": "myCustomHeaderValue"
    },
    "metadata": {
        "metadata-key-1": "metadata-value-1",
        "metadata-key-2": "metadata-value-2"
    },
    "authorization": {
        "type": "oauth2",
        "clientId": "myclientId",
        "clientSecret": "myclientSecret",
        "provider": "http://authorization.com/oauth/token",
        "scope": "READ",
        "mode": "header"
    }
```
}
With the current Agent SDK, subscription schema builder only manage regular (and enum) string values (of json-schema) which doesn't allow us to manage our model. 

But json-schema support other types like: number, array, object,... So, we have added the support of these missing types.
We tried to minimize the impact on the existing builder API, but depending on when the IsString() method is called in the property builder it may introduce a little breaking change on other discovery agents: The IsString() must be called after "common" properties methods (see documentation update in this PR). 

Impacted applicative code:
```
apic.NewSubscriptionSchemaPropertyBuilder().
            SetName(appNameKey).
            IsString().
            SetDescription(subscriptionUsagePlanDescription).
            SetEnumValues(values)
            SetRequired()
```
Fixed applicative code:
```
apic.NewSubscriptionSchemaPropertyBuilder().
            SetName(appNameKey).
            SetDescription(subscriptionUsagePlanDescription).
            SetRequired().
            IsString().
                SetEnumValues(values)
```
We have identified this issue on the following projects:

- v7_discovery_agent/-/blob/master/pkg/agent/subscriptionschema.go
- azure_discovery_agent/-/blob/master/pkg/agent/subscriptions.go
- aws_apigw_discovery_agent/-/blob/master/pkg/agent/agent.go


mynecker:
@jcollins-axway thanks for pointing to the PR. In case of any breaking changes it should be not a big deal to adapt solace agent accordingly.

PR #310, opened 2022-02-03
APIGOV-54321 - fall back to using external apiid if primary key does not resolver service

jcollins-axway:


dfeldick (approved):

pkg/apic/apiservice.go
```diff
@@ -130,7 +130,10 @@ func (c *ServiceClient) getAPIServiceByPrimaryKey(primaryKey string) (*v1alpha1.
 
 func (c *ServiceClient) getAPIServiceFromCache(serviceBody *ServiceBody) (*v1alpha1.APIService, error) {
 	if serviceBody.PrimaryKey != "" {
-		return c.getAPIServiceByPrimaryKey(serviceBody.PrimaryKey)
+		apiService, err := c.getAPIServiceByPrimaryKey(serviceBody.PrimaryKey)
+		if apiService != nil && err == nil {
+			return apiService, err
+		}
```
sbolosan, line 136: do you not care if the get by primary key did return an error and at least log a warn?  like you tried but it failed so you used apiid?

jcollins-axway: this is only for determining how to process the service, either create or update.  I do not think we need to log that something was not in the cache.

sbolosan: ah, that makes sense...  ok, ignore.

sbolosan (approved):

tsjohns9 (approved):

PR #315, opened 2022-02-08
APIGOV-21451 - let api server generate a name

jcollins-axway:


pkg/apic/apiservice.go
```diff
@@ -101,10 +95,7 @@ func (c *ServiceClient) processService(serviceBody *ServiceBody) (*v1alpha1.APIS
 	if err != nil {
 		return nil, err
 	}
-	_, err = c.apiServiceDeployAPI(httpMethod, serviceURL, buffer)
-	if err == nil {
-		serviceBody.serviceContext.serviceName = serviceName
-	}
+	serviceBody.serviceContext.serviceName, err = c.apiServiceDeployAPI(httpMethod, serviceURL, buffer)
```
vivekschauhan, line 98: Seems like there will be a problem with scenario where a API is published and then the API is updated on data plane. The apiService object when the API is published for the first time will be cached without name property. On the call for updating the API the apiService (without Name property) is fetched from cache to prepare the sreviceURL (line:86) which will fail when placing the PUT call.

For gRPC mode it will be OK, since the cache will be updated based on the received CREATED event triggered by the POST call

pkg/apic/apiserver/models/api/v1/resourcemeta.go
```diff
@@ -18,7 +18,7 @@ type Meta interface {
 // ResourceMeta metadata for a ResourceInstance
 type ResourceMeta struct {
 	GroupVersionKind
-	Name  string `json:"name"`
+	Name  string `json:"name,omitempty"`
```
tsjohns9, line 21: Can you add/update a test in resourcemeta_test.go for a resource without a name?

tsjohns9 (approved):

pkg/apic/apiserver/models/api/v1/resourcemeta.go
jcollins-axway: will do

vivekschauhan (approved):

PR #317, opened 2022-02-08
APIGOV-21092 Add team ID to metric event

sagarkal:


pkg/transaction/eventgenerator.go
```diff
@@ -78,13 +78,15 @@ func (e *Generator) trackMetrics(summaryEvent LogEvent, bytes int64) {
 		if summaryEvent.TransactionSummary.Application != nil {
 			appName = summaryEvent.TransactionSummary.Application.Name
 		}
-		teamName := ""
+		var team metric.Team
```
jcollins-axway, line 81: the team object here is the same as in the transaction details, can we make a single type?

pkg/transaction/metric/cachestorage.go
```diff
@@ -147,7 +147,7 @@ func (c *cacheStorage) loadAPIMetric(storageCache cache.Cache) {
 
 			var apiStatusMetric *APIMetric
 			for _, duration := range apiMetric.Values {
-				apiStatusMetric = c.collector.updateMetric(APIDetails{apiMetric.API.ID, apiMetric.API.Name, apiMetric.API.Revision}, apiMetric.StatusCode, duration)
+				apiStatusMetric = c.collector.updateMetric(APIDetails{apiMetric.API.ID, apiMetric.API.Name, apiMetric.API.Revision}, apiMetric.StatusCode, duration, apiMetric.Team)
```
jcollins-axway, line 150: since this api will always be part of the same team, maybe put the team details in API Details object, as to not change the signature of this method

sagarkal: well what if the API ownership changes? I know that its not possible in v7, but not sure about other gateways

jcollins-axway: but it can be part of the APIDetails object that we are creating here, should not matter if it changes.

sagarkal: yeah you're right, I realized it later and deleted my comment but then it was too late :)

jcollins-axway: `apiStatusMetric = c.collector.updateMetric(APIDetails{apiMetric.API.ID, apiMetric.API.Name, apiMetric.API.Revision, apiMetric.Team}, apiMetric.StatusCode, duration`

`type APIDetails struct {
	ID       string `json:"id"`
	Name     string `json:"name"`
	Revision int    `json:"revision"`
        TeamID string `json:"team"`
}`

sagarkal: Team ID can be extracted from API object 
`apiStatusMetric = c.collector.updateMetric(APIDetails{apiMetric.API.ID, apiMetric.API.Name, apiMetric.API.Revision, apiMetric.API.TeamID}, apiMetric.StatusCode, duration)`

tsjohns9 (approved):

kflan-io (approved):

jcollins-axway (approved):

vivekschauhan (approved):

PR #319, opened 2022-02-09
APIGOV-21642 - return metadata for cache job as that is the key used in the cache

jcollins-axway:


pkg/agent/discoverycache.go
```diff
@@ -75,7 +75,7 @@ func (j *discoveryCache) Execute() error {
 	log.Trace("executing API cache update job")
 	j.updateAPICache()
 	if agent.cfg.GetAgentType() == config.DiscoveryAgent {
-		j.updatePIServiceInstancesCache()
+		j.updateAPIServiceInstancesCache()
```
sbolosan, line 78: lol, how did we miss this? :)

sbolosan (approved):

PR #322, opened 2022-02-11
APIGOV-21519 Move SDK technical attributes to x-agent-details subresource

tsjohns9:
- Create a new subresource called 'x-agent-details' for API Services, Instances, Revisions, and Consumer Instances to hold the SDK technical attributes
- Add functions to easily get, add, and update the x-agent-details subresource.
- Moved the technical attributes, externalAPIID, externalAPIPrimaryKey, externalAPIName, away from the attributes field, and onto the x-agent-details subresource (except consumer instance which still has all the old attributes and the subresource).
- Provided methods to the ServiceBodyBuilder for agents to add additional details to the x-agent-details subresource.
- Update the models to remove subresources that are part of the resource definition


tsjohns9:
Ignore all the model updates. The main changes are found in `agent`, `util`,  and the top of `apic`

pkg/agent/discoverycache.go
```diff
@@ -20,7 +22,7 @@ import (
 const (
 	apiServerPageSize   = 100
 	healthcheckEndpoint = "central"
-	apiServerFields     = "metadata,name,title,owner,attributes"
+	apiServerFields     = "metadata,name,title,owner,attributes,x-agent-details"
```
jcollins-axway, line 25: do we still need/want attributes?

pkg/apic/apiservicerevision.go
```diff
@@ -298,7 +312,32 @@ func (c *ServiceClient) GetAPIRevisionByName(revisionName string) (*v1alpha1.API
...
+
+func getRevisionPrefix(serviceBody *ServiceBody) string {
+	if serviceBody.Stage != "" {
+		return sanitizeAPIName(fmt.Sprintf("%s-%s", serviceBody.serviceContext.serviceName, serviceBody.Stage))
+	}
+	return sanitizeAPIName(serviceBody.serviceContext.serviceName)
+}
+
+// getRevisionDefinitionType -
+func getRevisionDefinitionType(serviceBody ServiceBody) string {
```
jcollins-axway, line 338: maybe this is better put as a function on ServiceBody?

pkg/agent/discoverycache.go
tsjohns9: We still need it for consumer instances. Consumer instances save all the attributes, and also save the same attributes in the x-agent-details. Once we can stop querying for consumer instances on the api id, then we can probably remove the attributes.

pkg/apic/apiservicerevision.go
tsjohns9: I'll look into it

tsjohns9:
We can wait until the release is done.

sbolosan (approved):

PR #329, opened 2022-02-19
APIGOV-21763 add type to perform attribute migration

tsjohns9:
- Added a new package to perform attribute the migration.
- Migration will be called by the discovery cache upon finding API Services. If the 'x-agent-details' sub resource is not found, then the migration will be performed for that service.
- Fetch ConsumerInstances by metadata reference to service instead of externalAPIID
- Add public funcs for agents to define their own attributes for migration

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -15,13 +15,13 @@ import (
 	"github.com/Axway/agent-sdk/pkg/agent/resource"
 	"github.com/Axway/agent-sdk/pkg/agent/stream"
 	"github.com/Axway/agent-sdk/pkg/api"
-
 	"github.com/Axway/agent-sdk/pkg/apic"
 	apiV1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
 	"github.com/Axway/agent-sdk/pkg/apic/auth"
 	"github.com/Axway/agent-sdk/pkg/cache"
 	"github.com/Axway/agent-sdk/pkg/config"
 	"github.com/Axway/agent-sdk/pkg/jobs"
+	"github.com/Axway/agent-sdk/pkg/migrate"
 	"github.com/Axway/agent-sdk/pkg/util"
 	"github.com/Axway/agent-sdk/pkg/util/errors"
 	hc "github.com/Axway/agent-sdk/pkg/util/healthcheck"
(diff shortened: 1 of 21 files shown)


pkg/apic/service.go
```diff
@@ -120,11 +121,12 @@ func (c *ServiceClient) DeleteAPIServiceInstance(name string) error {
...
 // GetConsumerInstanceByID -
 func (c *ServiceClient) GetConsumerInstanceByID(consumerInstanceID string) (*v1alpha1.ConsumerInstance, error) {
-	return c.getConsumerInstanceByID((consumerInstanceID))
+	return c.getConsumerInstanceByID(consumerInstanceID)
 }
 
-// GetConsumerInstancesByExternalAPIID -
+// GetConsumerInstancesByExternalAPIID - DEPRECATED
 func (c *ServiceClient) GetConsumerInstancesByExternalAPIID(externalAPIID string) ([]*v1alpha1.ConsumerInstance, error) {
+	log.DeprecationWarningReplace("GetConsumerInstancesByExternalAPIID", "")
```
tsjohns9, line 129: @jcollins-axway This should be deprecated, right?

pkg/agent/agent.go
```diff
@@ -250,7 +252,7 @@ func startAPIServiceCache() {
 			return
 		}
 		// Start the full update after the first interval
-		go startDiscoveryCache(agent.instanceCacheLock)
+		go startDiscoveryCache(agent.instanceCacheLock, am)
 		log.Tracef("registered API cache update job: %s", id)
 	} else {
 		// Load cache from API initially. Following updates to cache will be done using watch events
```
tsjohns9, line 258: So, I changed the migration according to Jason's suggestions so that we didn't have an issue with some agents not having the agent resource. Now I realized that this isn't going to work for stream mode because we don't update the discovery cache with new resources when the agent starts up. This relates back to a defect I recently made where the resources won't get made if the sdk cache is outdated with the central environment, APIGOV-21724.

vivekschauhan: @tsjohns9, with streaming mode when the persistent cache is turned on (by default turned off) we do not skip discovery rather we do not run the initialization for the cache by making the API call since we already have them loaded from the persistent store. To run the migration we should iterate over the cached api-service to perform the migration.

Also may be we should re-consider running the migration logic as part of the job which might have some race condition since we are adding more processing time with the migration logic, I understand it is one time as Migrate() method checks for x-agent-details subresource but if we have it running as part of job, other jobs might get kicked off while the discovery cache job is processing the migration and try to look for items in cache, so in my opinion the migration should be executed first and then we should start to initialize jobs as part of agent initialization, may be do a explicit call to Execute() method before starting the job like we do before starting the streaming job to update the cache in persistent cache is not available.

pkg/agent/discoverycache.go
```diff
@@ -105,6 +110,10 @@ func (j *discoveryCache) updateAPICache() {
 	)
 
 	for _, svc := range apiServices {
+		svc, err := j.migrator.Migrate(svc)
+		if err != nil {
+			panic(fmt.Errorf("failed to migrate attributes: %s", err))
```
jcollins-axway, line 115: Do we want to panic or just throw an error?

PR #331, opened 2022-02-22
APIGOV-21763 Attribute migration

tsjohns9:


pkg/agent/agent.go
```diff
@@ -146,13 +147,29 @@ func InitializeWithAgentFeatures(centralCfg config.CentralConfig, agentFeaturesC
...
 		if util.IsNotTest() && !isRunningInDockerContainer() {
 			hc.StartPeriodicHealthCheck()
 		}
 
 		if util.IsNotTest() && agent.agentFeaturesCfg.ConnectionToCentralEnabled() {
 			StartAgentStatusUpdate()
+			migration := migrate.NewAttributeMigration(agent.apicClient, agent.cfg)
+			// register the update cache job
+			discoveryCache := newDiscoveryCache(agent.agentResourceManager, false, agent.instanceCacheLock)
+			discoveryCache.Execute()
```
tsjohns9, line 160: @vivekschauhan I moved the call to populate the cache out so that it blocks the other jobs until this is finished. This fixes the race condition. But, I am still unsure if this is what you are looking for because you also mentioned looping through persistent cache in stream mode. This way will populate the cache from central regardless if it is stream or poll mode which makes the persistent cache not that useful.

pkg/agent/agent.go
```diff
@@ -146,13 +147,29 @@ func InitializeWithAgentFeatures(centralCfg config.CentralConfig, agentFeaturesC
...
+
+			for _, key := range agent.cacheManager.GetAPIServiceCache().GetKeys() {
```
tsjohns9, line 162: The looping here is temporary until I have a response to my question above.

pkg/agent/discoverycache.go
```diff
@@ -105,6 +110,9 @@ func (j *discoveryCache) updateAPICache() {
 	)
 
 	for _, svc := range apiServices {
+		if j.migrator != nil {
+			svc, _ = j.migrator.Migrate(svc)
```
tsjohns9, line 114: I found leaving the migrator in the discovery cache to be the most efficient way to fetch a resource, update it, then save it to the cache. @vivekschauhan Let me know if you are ok with this, or if you have another idea based on how you want to populate the initial cache.

pkg/agent/agent.go
vivekschauhan: @tsjohns9, for streaming mode what I meant was instead of updating the cache by making API call and executing the migration logic, we should iterate over the cached APIServices and run migration on them.

pkg/agent/discoverycache.go
vivekschauhan: Looks fine except iterating over the cached APIService(from persisted cache if available) in streaming mode

pkg/migrate/attributemigration.go
```diff
@@ -0,0 +1,299 @@
...
+	defs.AttrExternalAPIStage,
+	defs.AttrCreatedBy,
+}
+
+// AttrMigrator interface for performing an attribute migration
+type AttrMigrator interface {
+	Migrate(ri *v1.ResourceInstance) (*v1.ResourceInstance, error)
+}
+
+var regexes = make([]string, 0)
```
jcollins-axway, line 33: Do we still want to allow regexes in this? It could be dangerous to allow an agent to set that and we move more than necessary.

pkg/apic/apiserviceinstance.go
```diff
@@ -36,7 +36,7 @@ func (c *ServiceClient) buildAPIServiceInstance(
 			GroupVersionKind: mv1a.APIServiceInstanceGVK(),
 			Name:             name,
 			Title:            serviceBody.NameToPush,
-			Attributes:       util.MergeMapStringString(map[string]string{}, serviceBody.InstanceAttributes),
+			Attributes:       util.MergeMapStringString(serviceBody.InstanceAttributes),
```
jcollins-axway, line 39: Only one map is being sent in, does it still need to be sent to mergemap? Same with the other places

pkg/apic/client.go
```diff
@@ -683,27 +686,29 @@ func (c *ServiceClient) CreateSubResourceScoped(
...
 
 	for subName, sub := range subs {
-		wg.Add(1)
+		if strings.HasPrefix(subName, "x-") {
+			wg.Add(1)
 
-		base := c.cfg.GetURL()
-		url := fmt.Sprintf("%s/apis/%s/%s/%s/%s/%s/%s/%s", base, group, version, scopeKindPlural, scopeName, resKindPlural, name, subName)
+			base := c.cfg.GetURL()
+			url := fmt.Sprintf("%s/apis/%s/%s/%s/%s/%s/%s/%s", base, group, version, scopeKindPlural, scopeName, resKindPlural, name, subName)
```
jcollins-axway, line 693: Not that it is needed for this story, but wonder if we can utilize the registry in the apiserver/v1 package to assist in creating this URL

jcollins-axway (approved):

pkg/apic/apiserviceinstance.go
tsjohns9: If the map being passed in is nil, then it will return a new map, so it avoids a nil check

pkg/apic/client.go
tsjohns9: That would be interesting. That was made for the resource specific clients. Not sure what it would take to start using that.

pkg/migrate/attributemigration.go
tsjohns9: I think we need it. The AWS discovery agent has this line
```
serviceBody.Attributes[serviceBody.Stage+"Hash"] = convertUnitToString(serviceBodyHash)
```

Since the attribute is not a constant I figured we needed a regex to match this one.

pkg/migrate/attributemigration.go
```diff
@@ -0,0 +1,301 @@
...
+		}
+	}
+
+	util.SetAgentDetails(ri, details)
+
+	return item
+}
+
+func queryFunc(name string) string {
+	return fmt.Sprintf("metadata.references.name==%s", name)
```
tsjohns9, line 300: @vivekschauhan @jcollins-axway I found a bug. This query does not work for api service instances. It is the only resource that does not have a reference to the API Service. It only has a reference to the revision it was made for. The revision and consumer instances both have a reference to the API Service. Can we update the discovery process to include a ref to the API Service when creating the APIServiceInstance?

pkg/migrate/attributemigration.go
```diff
@@ -0,0 +1,336 @@
...
+// updateInst gets a list of instances for the service and updates their attributes.
+func (m *AttributeMigration) updateInst(ri *v1.ResourceInstance) error {
+	revURL := m.cfg.GetRevisionsURL()
+
+	q := map[string]string{
+		"query":  queryFunc(ri.Name),
+		"fields": "name",
+	}
+
+	revs, err := m.client.GetAPIV1ResourceInstancesWithPageSize(q, revURL, 100)
```
tsjohns9, line 155: I realized that the query that was being used was incorrect for api service instances. They only have a ref to a revision, so I updated this func to fetch all revisions for the service, and then update each instance for each revision.

pkg/migrate/attributemigration.go
jcollins-axway: I forgot about that one, good call

pkg/apic/apiserviceinstance.go
jcollins-axway: That makes sense. However if there is only 1 map we are using extra cycles to iterate over all items. Maybe throw a length check in the MergeMapStringString func, to return the map if only 1?

vivekschauhan (approved):

jcollins-axway (approved):

pkg/apic/apiserviceinstance.go
tsjohns9: I'll make new func to check for nil maps

PR #334, opened 2022-02-28
APIGOV-21855 - new provisioning package and access request handler

jcollins-axway:


sbolosan (approved):
minus the lint stuff

pkg/agent/handler/apiservicehandler.go
```diff
@@ -0,0 +1,44 @@
...
+		agentCacheManager: agentCacheManager,
+	}
+}
+
+func (h *apiSvcHandler) Handle(action proto.Event_Type, _ *proto.EventMeta, resource *v1.ResourceInstance) error {
+	if resource.Kind != apiService {
+		return nil
+	}
+
+	id, ok := resource.Attributes[definitions.AttrExternalAPIID]
```
tsjohns9, line 30: Keep in mind this will need to be updated to read from the sub resource.

pkg/apic/provisioning/credentials.go
```diff
@@ -0,0 +1,98 @@
...
+	id     string
+	secret string
+}
+
+type apiKeyCredential struct {
+	key string
+}
+
+// CredentialBuilder - builder to create new credentials to send to Central
+type CredentialBuilder interface {
```
tsjohns9, line 24: Can a credential have both oauth and api key defined?

tsjohns9: I'm hesitant to have a single builder for all credentials. If more credentials are added then we have to keep growing this interface and the types that implement it. Can we some kind of `Credential` interface for other hypothetical structs like `OAuthCredential` `APIKeyCredential` to implement?

jcollins-axway: There would be separate requests for each type, but this builder will support all credential types. If they try to set 2 different types an error is returned

pkg/agent/handler/apiservicehandler.go
jcollins-axway: Sure thing, I just broke this out from the single handlers file, will merge those changes when they are in main

pkg/apic/provisioning/credentials.go
tsjohns9: Based on the tests it doesn't look like we will be setting both types of credentials at once, so I don't think a builder is the right approach here.

tsjohns9: Something like this to break them up and keep each credential independent of the other

```
type credentialType int

const (
	// credentialTypeOAuth - OAuth credentials
	credentialTypeOAuth credentialType = iota + 1
	// credentialTypeAPIKey - APIKey credentials
	credentialTypeAPIKey
)

type Credential struct {
	credentialType credentialType
	data           interface{}
	reference      string
}

type oAuthCredential struct {
	id     string
	secret string
}

type apiKeyCredential struct {
	key string
}

func NewOAuthCredential(id, secret, ref string) Credential {
	c := Credential{
		credentialType: credentialTypeOAuth,
		reference:      ref,
	}
	c.data = &oAuthCredential{
		id:     id,
		secret: secret,
	}

	return c
}

func NewAPIKeyCredential(key, ref string) Credential {
	c := Credential{
		credentialType: credentialTypeAPIKey,
		reference:      ref,
	}
	c.data = &apiKeyCredential{
		key: key,
	}

	return c
}
```

pkg/apic/provisioning/requeststatus.go
```diff
@@ -0,0 +1,69 @@
...
+const statusSetError = "can not set status as it's already been set"
+
+// RequestStatus - holds info about the status of the request
+type RequestStatus struct {
+	status  status
+	message string
+}
+
+// RequestStatusBuilder - builder to create new request status
+type RequestStatusBuilder interface {
```
tsjohns9, line 14: I have the same feedback on this builder too. But, I don't see enough of how it is meant to be used from the tests to know how it is meant to be used.

pkg/agent/handler/apiservicehandler.go
tsjohns9: I updated it on the migration branch, which will probably be merged before this so it should be fine.

pkg/apic/provisioning/credentials.go
tsjohns9: How will the agents be creating a credential? I see it will return an error, but I think the builder adds more complexity than having two separate types.

pkg/apic/provisioning/requeststatus.go
tsjohns9:
```
const (
	// Success - request was successful
	Success status = iota + 1
	// Failed - request failed
	Failed
)

type RequestStatus struct {
	status  status
	message string
}

func NewSuccessStatus() RequestStatus {
	return RequestStatus{
		status:  Success,
		message: "",
	}
}

func NewErrorStatus(message string) RequestStatus {
	return RequestStatus{
		status:  Failed,
		message: message,
	}
}
```

tsjohns9: This avoids error checking and a processing func that is needed with the builder.

jcollins-axway: The agents will use both builders to create their respective types.  These types will be used as return data for the provision or deprovision method requests

pkg/apic/provisioning/credentials.go
```diff
@@ -0,0 +1,98 @@
+package provisioning
+
+import "fmt"
+
+const credentialTypeSetError = "can not set credential as %s as its already set as another %s"
+
+// Credential - holds the details about the credential to send to encrypt and send to platform
+type Credential struct {
```
vivekschauhan, line 8: This should probably be an interface and an internal struct should hold the values that gets populated by the builder and builder returns to the provisioning interface so that the callback can return it back to agent SDK

pkg/apic/provisioning/requeststatus.go
```diff
@@ -0,0 +1,69 @@
+package provisioning
+
+import "fmt"
+
+const statusSetError = "can not set status as it's already been set"
+
+// RequestStatus - holds info about the status of the request
+type RequestStatus struct {
```
vivekschauhan, line 8: Again this should be an interface with internal struct populated by the builder

pkg/apic/provisioning/requeststatus.go
```diff
@@ -0,0 +1,69 @@
...
+	status  status
+	message string
+}
+
+// RequestStatusBuilder - builder to create new request status
+type RequestStatusBuilder interface {
+	Success() RequestStatusBuilder
+	Failed(message string) RequestStatusBuilder
+	Process() (*RequestStatus, error)
```
vivekschauhan, line 17: Returning a pointer here however the provisioning interface are returning an object

pkg/apic/provisioning/requeststatus.go
```diff
@@ -0,0 +1,69 @@
...
+
+import "fmt"
+
+const statusSetError = "can not set status as it's already been set"
+
+// RequestStatus - holds info about the status of the request
+type RequestStatus struct {
+	status  status
+	message string
+}
```
vivekschauhan, line 11: May be we can include the reference here, the caller of the provisioning interface methods can use it and associate it back to the resource (ManagedApplication, AccessRequest, Credential).

pkg/apic/provisioning/requeststatus.go
```diff
@@ -0,0 +1,69 @@
...
+
+// NewRequestStatusBuilder - create a request status builder
+func NewRequestStatusBuilder() RequestStatusBuilder {
+	return &requestStatusBuilder{
+		status: &RequestStatus{},
+	}
+}
+
+// Process - process the builder, returning errors
+func (r *requestStatusBuilder) Process() (*RequestStatus, error) {
```
vivekschauhan, line 33: Do we need this, we could directly return the status from Success or Failed methods. Also not sure of we need to report error. Trying to see how the agents will call it and it does not seem right

```
func (a *Subscription) ApplicationRequestDeprovision(applicationRequest provisioning.ApplicationRequest) (status provisioning.RequestStatus) {
    ...
    status, err := builder.Success().Process()
    if err != nil {
        // what should the agent do ?
    }
    return status
}
```

Would it be much easier for the agent to call 
```
func (a *Subscription) ApplicationRequestDeprovision(applicationRequest provisioning.ApplicationRequest) (status provisioning.RequestStatus) {
    ...
    return builder.Success()
}
```

pkg/apic/provisioning/applicationrequest.go
```diff
@@ -0,0 +1,6 @@
+package provisioning
+
+// ApplicationRequest - interface for agents to use to get necessary application request details
+type ApplicationRequest interface {
+	GetApplicationName() string // returns the name of the managed application for this credential
```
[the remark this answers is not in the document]
jcollins-axway: Added GetProperties() to all of these

pkg/apic/provisioning/credentials.go
jcollins-axway: Made this private and created a Credential interface above it

pkg/apic/provisioning/requeststatus.go
jcollins-axway: same

pkg/apic/provisioning/requeststatus.go
jcollins-axway: Removed this

pkg/apic/provisioning/requeststatus.go
jcollins-axway: Changed it, removed process method. Return the RequestStatus interface directly

pkg/apic/provisioning/requeststatus.go
jcollins-axway: After Vivek's feedback and subsequent updates this may be better shown.  The credential and requestStatus types are meant to be SDK only types.  So the agent implementation will not have direct access to their values.  The builders have been updated to return only interfaces and the concrete types are private now.

sbolosan (approved):

pkg/apic/provisioning/credentials.go
jcollins-axway: The builders allow us greater flexibility for changes in the SDK in the future. Thinking for this we may need to add additional setter methods for the API Key to say if it's in the headers or query params. With the builder we can add that additional method without causing issues on the agent implementation.

vivekschauhan (approved):

PR #336, opened 2022-03-03
APIGOV-21854 GRPC event handler for AccessRequests

tsjohns9:
First round of gathering feedback for processing grpc events for AccessRequests. The main changes are in the agent/handler/accessrequest.go file. I had to pull in the migration branch, so all those changes are showing up in the review too.


pkg/agent/handler/accessrequest.go
```diff
@@ -0,0 +1,188 @@
+package handler
```
tsjohns9, line 1: Bulk of the work is here

pkg/apic/provisioning/definitions.go
```diff
@@ -2,20 +2,20 @@ package provisioning
 
 // enums
 
-// credentialType - the type fo credential
```
tsjohns9, line 5: Some changes across several files in the provisioning package too

sbolosan (approved):

pkg/agent/stream/watchtopic.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 // getOrCreateWatchTopic attempts to retrieve a watch topic from central, or create one if it does not exist.
-func getOrCreateWatchTopic(name, scope string, rc ResourceClient, agentType config.AgentType) (*mv1.WatchTopic, error) {
```
tsjohns9, line 16: Changes in this file for creating the watch topic for the new resources.

pkg/agent/handler/accessrequest.go
```diff
@@ -0,0 +1,188 @@
...
+	v1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
+	mv1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/management/v1alpha1"
+	defs "github.com/Axway/agent-sdk/pkg/apic/definitions"
+	prov "github.com/Axway/agent-sdk/pkg/apic/provisioning"
+	"github.com/Axway/agent-sdk/pkg/util"
+	"github.com/Axway/agent-sdk/pkg/watchmanager/proto"
+)
+
+const (
+	provision     = "Provision"
```
tsjohns9, line 17: We need a place to put all these constants. This is a temporary spot.

vivekschauhan (commented):
Unrelated by I don't see a PR yet for the technical attribute migration changes, most of the changes here are on top of those changes. I know even Shane's changes are based on top of those changes. May be we should get that attribute change merged and then rebase these new branches

pkg/agent/handler/accessrequest.go
```diff
@@ -0,0 +1,188 @@
...
+func NewAccessRequestHandler(prov arProvisioner, cache agentcache.Manager, client client) Handler {
+	return &accessRequestHandler{
+		prov:   prov,
+		cache:  cache,
+		client: client,
+	}
+}
+
+func (h *accessRequestHandler) Handle(action proto.Event_Type, _ *proto.EventMeta, resource *v1.ResourceInstance) error {
+	if resource.Kind != mv1.AccessRequestGVK().Kind {
```
vivekschauhan, line 45: We probably need a check to allow processing if provisioner interface is set(will be set only for DiscoveryAgent type)

pkg/agent/handler/accessrequest.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return err
+	}
+
+	req, err := h.newReq(ar, util.GetAgentDetails(app))
+	if err != nil {
+		return err
+	}
+
+	if ar.Status == nil || ar.Status.Level == "" {
```
vivekschauhan, line 69: This should be move up in processing

pkg/agent/handler/accessrequest.go
```diff
@@ -0,0 +1,188 @@
...
+
+	details := util.MergeMapStringInterface(util.GetAgentDetails(ar), status.GetProperties())
+	util.SetAgentDetails(ar, details)
+
+	bts, err := json.Marshal(ar)
+	if err != nil {
+		return err
+	}
+
+	_, err = h.client.UpdateResource(ar.Metadata.SelfLink, bts)
```
vivekschauhan, line 99: Since we are not making any change to the spec just the subresources(x-agent-detail and status) are getting updated, we probably do not need the update call

pkg/agent/handler/accessrequest.go
```diff
@@ -0,0 +1,188 @@
...
+
+	err = h.client.CreateSubResourceScoped(
+		mv1.EnvironmentResourceName,
+		ar.Metadata.Scope.Name,
+		ar.PluralName(),
+		ar.Name,
+		ar.Group,
+		ar.APIVersion,
+		map[string]interface{}{
+			defs.XAgentDetails: util.GetAgentDetails(ar),
```
vivekschauhan, line 112: We should add status subresource here since we are updating it

pkg/apic/provisioning/accessrequest.go
```diff
@@ -2,7 +2,12 @@ package provisioning
 
 // AccessRequest - interface for agents to use to get necessary access request details
 type AccessRequest interface {
-	GetApplicationName() string         // returns the name of the managed application for this credential
-	GetAPIID() string                   // returns the reference of the API on the data plane to be used
-	GetProperty(key string) interface{} // returns the property value for the key sent in
+	// GetApplicationName returns the name of the managed application for this credential
+	GetApplicationName() string
+	// GetAPIID returns the reference of the API on the data plane to be used
+	GetAPIID() string
```
vivekschauhan, line 8: May be we should add interface for getting the externalAPIStage as well. We probably do not need externalPrimaryKey though

pkg/apic/provisioning/definitions.go
```diff
@@ -29,16 +29,30 @@ const (
...
-	Success status = iota + 1
+	Success Status = iota + 1
 	// Failed - request failed
 	Failed
 )
 
+func (c Status) String() string {
+	return map[Status]string{
+		Success: "success",
+		Failed:  "failed",
```
vivekschauhan, line 52: These constants are getting used for building the status subresource, so the value should correspond to "Success" and "Error"(instead of "failed" for error case)

pkg/apic/provisioning/requeststatus.go
```diff
@@ -1,63 +1,112 @@
...
+		rsr := apiv1.ResourceStatusReason{
+			Type:      r.GetStatus().String(),
+			Detail:    msg,
+			Timestamp: apiv1.Time{},
+		}
+		reasons = append(reasons, rsr)
+	}
+
+	return apiv1.ResourceStatus{
+		Level:   msg,
```
vivekschauhan, line 109: This should be r.GetStatus().String() as well.

pkg/agent/handler/accessrequest.go
tsjohns9: If there is no processor for a TA, then what will we be doing in here?

pkg/agent/handler/accessrequest.go
tsjohns9: The Status field needs to be updated too right?

pkg/agent/handler/accessrequest.go
tsjohns9: Does the status get updated by calling the subresource endpoint?

PR #339, opened 2022-03-04
APIGOV-21860 - register credential request defs and assign to api service instance

jcollins-axway:
- create builder
- use builder to register
- based on auth profile assign a credential request def


pkg/agent/instancevalidator.go
```diff
@@ -96,22 +96,23 @@ func (j *instanceValidator) getServiceInstanceCount(attName, attValue string) in
...
+	msg := "Deleted catalog item %s from Amplify Central"
 
-	defer func() {
-		// remove the api service instance from the cache for both scenarios
-		if j.isAgentPollMode {
-			// In GRPC mode delete is done on receiving delete event from service
-			agent.cacheManager.DeleteAPIServiceInstance(ri.Metadata.ID)
-		}
-	}()
+	err := agent.apicClient.DeleteAPIServiceInstance(ri.Name)
```
vivekschauhan, line 103: May we should run the DeleteAPIs automation test to avoid any regressions.

pkg/agent/instancevalidator.go
```diff
@@ -96,22 +96,23 @@ func (j *instanceValidator) getServiceInstanceCount(attName, attValue string) in
...
+	if err != nil {
+		log.Error(utilErrors.Wrap(ErrDeletingCatalogItem, err.Error()).FormatError(ri.Title))
+		return
+	}
+	if j.isAgentPollMode {
```
vivekschauhan, line 108: Seems it would cause issue in gRPC mode where the validator deletes the instance and the item is not removed from the cache so shouldDeleteService is not going to remove the APIService. When the handler will receive the delete resource event, the instance will be removed from the cache by handler, however on next validator execution the above DeleteAPIServiceInstance call is going to fail since it will not find the instance and the service will not be removed.

May be we can remove the isAgentPollMode check here

pkg/agent/provisioning.go
```diff
@@ -0,0 +1,67 @@
...
+// credential request definitions
+
+// createOrUpdateCredentialRequestDefinition -
+func createOrUpdateCredentialRequestDefinition(data *v1alpha1.CredentialRequestDefinition) (*v1alpha1.CredentialRequestDefinition, error) {
+	// TODO - check cache for credential request, update if needed
+	return agent.apicClient.RegisterCredentialRequestDefinition(data, true)
+}
+
+// NewCredentialRequestBuilder - called by the agents to build and register a new credential reqest definition
+func NewCredentialRequestBuilder() provisioning.CredentialRequestBuilder {
```
vivekschauhan, line 17: Can we change the name to NewCredentialRequestDefinitionBuilder() or may be short NewCRDBuilder() to avoid having long name?

pkg/apic/apiserviceinstance.go
```diff
@@ -71,8 +83,62 @@ func (c *ServiceClient) updateAPIServiceInstance(
...
+	newARD := &mv1a.AccessRequestDefinition{}
+	err = json.Unmarshal(response, newARD)
+	if err != nil {
+		return nil, err
+	}
+
+	return newARD, nil
+}
+
+func (c *ServiceClient) createAccessRequestDefintion(serviceBody *ServiceBody) error {
```
vivekschauhan, line 109: Should this be an internal processing in ServiceBody?

pkg/apic/apiserviceinstance.go
```diff
@@ -71,8 +83,62 @@ func (c *ServiceClient) updateAPIServiceInstance(
...
+	oauthScopes := make([]string, 0)
+	for scope := range serviceBody.GetScopes() {
+		oauthScopes = append(oauthScopes, scope)
+	}
+	if len(oauthScopes) > 0 {
+		newARD, err := provisioning.NewAccessRequestBuilder(c.createOrUpdateAccessRequestDefinition).
```
vivekschauhan, line 115: We probably need a way to avoid making API call if the definition is same

vivekschauhan (approved):

pkg/agent/instancevalidator.go
jcollins-axway: Removed, will turn on grpc in my env to test this today

pkg/agent/instancevalidator.go
jcollins-axway: Will ask QA to run some tests

pkg/apic/apiserviceinstance.go
jcollins-axway: Moved, called as part of the builder

pkg/apic/apiserviceinstance.go
jcollins-axway: That is a todo, yes

pkg/agent/provisioning.go
tsjohns9: same thing for the func above too

pkg/apic/client.go
```diff
@@ -764,3 +768,49 @@ func (c *ServiceClient) CreateResource(url string, bts []byte) (*apiv1.ResourceI
...
+	url := fmt.Sprintf(c.cfg.GetEnvironmentURL() + "/credentialrequestdefinitions")
+
+	response, err := c.ExecuteAPI(coreapi.POST, url, nil, crdBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	newCRD := &mv1a.CredentialRequestDefinition{}
+	err = json.Unmarshal(response, newCRD)
+	if err != nil {
```
tsjohns9, line 788: If you want to drop this if check, you can return `newCRD, err`

pkg/apic/provisioning/accessrequestdefinitionbuilder_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+			_, err := builder.Register()
+
+			if tt.wantErr {
+				assert.NotNil(t, err)
+				assert.False(t, registerFuncCalled)
+			} else {
+				assert.Nil(t, err)
+				assert.True(t, registerFuncCalled)
+				// assert.NotNil(t, builtDef)
```
tsjohns9, line 66: dead code

pkg/agent/provisioning.go
jcollins-axway: done

pkg/apic/provisioning/accessrequestdefinitionbuilder_test.go
jcollins-axway: part of my todo

tsjohns9 (approved):

pkg/apic/provisioning/schemabuilder_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewSchemaBuilder(t *testing.T) {
+	builder := NewSchemaBuilder()
+	assert.NotNil(t, builder)
+
+	schemaBuilderProps := builder.(*schemaBuilder)
```
tsjohns9, line 13: Can you check if the type conversion was good too please

tsjohns9: Nevermind, it's a test file.

PR #343, opened 2022-03-09
APIGOV-22010 - Fix for updating agent specific subresources

vivekschauhan:
- Changes to update only x-agent-details subresource rather than placing PUT call for all existing subresources on published resources.
- For APIService status is included for subresource updates on discovered APIServices


sbolosan (approved):
Thank you

PR #347, opened 2022-03-10
APIGOV-21858 - updates found while implementing agent

jcollins-axway:


pkg/agent/cache/manager.go
```diff
@@ -533,9 +533,11 @@ func (c *cacheManager) GetSequence(watchTopicName string) int64 {
 }
 
 func (c *cacheManager) ApplyResourceReadLock() {
+	log.Trace("LOCKED")
```
jcollins-axway, line 536: Will remove

pkg/agent/cache/manager.go
```diff
@@ -533,9 +533,11 @@ func (c *cacheManager) GetSequence(watchTopicName string) int64 {
...
 	c.resourceCacheReadLock.Lock()
 }
 
 func (c *cacheManager) ReleaseResourceReadLock() {
+	log.Trace("UNLOCKED")
```
jcollins-axway, line 541: Will remove

pkg/agent/stream/client.go
```diff
@@ -203,7 +203,7 @@ func getWatchServiceHostPort(cfg config.CentralConfig) (string, int) {
 
 // Start creates and starts everything needed for a stream connection to central.
 func (c *streamer) Start() error {
-	events, eventErrorCh := make(chan *proto.Event), make(chan error)
+	events, eventErrorCh := make(chan *proto.Event, 100), make(chan error, 100)
```
jcollins-axway, line 206: temporary work around for deadlock issue

pkg/agent/handler/accessrequest.go
```diff
@@ -82,16 +78,15 @@ func (h *accessRequestHandler) Handle(action proto.Event_Type, _ *proto.EventMet
 
 	var status prov.RequestStatus
 
-	if ar.Status.Level == statusPending && ar.State.Name == provision {
+	if ar.Status.Level == statusPending {
 		status = h.prov.AccessRequestProvision(req)
 	}
 
-	if ar.Status.Level == statusPending && ar.State.Name == deprovision {
+	if ar.Status.Level == statusPending {
```
vivekschauhan, line 85: This does not seem right, the pending status is being used for both provisioning and deprovisioning

jcollins-axway: The state is being remove, but this should be updated to check for DELETING state in resource metadata

pkg/agent/handler/handler.go
```diff
@@ -27,6 +27,11 @@ func isStatusFound(rs *v1.ResourceStatus) bool {
 	return true
 }
 
+func isNotStatusSubResourceUpdate(action proto.Event_Type, meta *proto.EventMeta) bool {
+	// TODO update this to check for change in status on resource meta
+	return (action == proto.Event_SUBRESOURCEUPDATED)
```
vivekschauhan, line 32: We probably need that meta.SubResource != "status" check here. I know we talked about "state" subresource, once that is added to the resources we need to add that check as well

vivekschauhan (approved):

pkg/agent/handler/handler.go
jcollins-axway: I will add that back

pkg/agent/handler/accessrequest.go
vivekschauhan: ar.Metadata.State == "DELETING", right ? Just confirming

pkg/agent/handler/managedapplication.go
```diff
@@ -31,8 +31,8 @@ func NewManagedApplicationHandler(prov managedAppProvision, cache agentcache.Man
 }
 
 // Handle processes grpc events triggered for ManagedApplications
-func (h *managedApplication) Handle(action proto.Event_Type, _ *proto.EventMeta, resource *v1.ResourceInstance) error {
-	if resource.Kind != mv1.ManagedApplicationGVK().Kind || h.prov == nil || action == proto.Event_SUBRESOURCEUPDATED {
+func (h *managedApplication) Handle(action proto.Event_Type, meta *proto.EventMeta, resource *v1.ResourceInstance) error {
+	if resource.Kind != mv1.ManagedApplicationGVK().Kind || h.prov == nil || isNotStatusSubResourceUpdate(action, meta) {
```
tsjohns9, line 35: How come you want to process some sub resource events? You should be getting an event for the AccessRequest when the sub resource is added to the main resource which makes processing sub resources redundant.

jcollins-axway: What about when a subresource is just updated, not added to a resource?

tsjohns9 (approved):

pkg/agent/handler/managedapplication.go
vivekschauhan: We will certainly get update event since the resource will be created by SaaS controller but we need to handle the sub resource update event as well since the resources can be created using API/CLI (without subresource) and then have an update call to update the subresource. 

Also, for the case where the agent fails the resource provisioning it is going to set the status as Error, but if someone wants to reprocess, they could update the subresource(there should be some sort of state transition validation though I believe that should done)

pkg/agent/handler/credential.go
```diff
@@ -98,30 +100,43 @@ func (h *credentials) Handle(action proto.Event_Type, _ *proto.EventMeta, resour
 		if err != nil {
```
vivekschauhan, line 100: We are probably missing to set the Error status on the credential resource, If the key is invalid (invalid PEM content) the newPub() is going to fail and return error however the status on the resource is not changed because the flow returns from this block, so it will remaining in Pending status.

jcollins-axway: made changes to catch this error and build a failed status

tsjohns9: We need to handle the failed status in the other handlers too

jcollins-axway: We will want to mark them as TODOs or create stories

pkg/agent/handler/managedapplication.go
tsjohns9: I still don't think it is necessary to handle a subresource event because we fetch the full resource each time we start processing events. So if there was an update to the status subresource we will still fetch the full AccessRequest which will then have all the sub resources, and then we will process it as a normal AccessRequest update.

vivekschauhan: Correct, we fetch the the full resource every time for all events (except for delete event), however SUBRESOURCEUPDATE is raised separately. If the resource itself is not modified but only the subresource is updated using PUT call to /resource/subresource, it will only raise SUBRESOURCEUPDATE event, there will be no UPADTE event for the resource itself.
So we need to handle the subresource event for those cases which is going to be the usual case when these subresoruces will be updated.

PR #349, opened 2022-03-11
APIGOV-20810 - updates for adding finalizers clean access request poller

jcollins-axway:


sbolosan (approved):

pkg/agent/handler/accessrequest.go
```diff
@@ -57,7 +58,7 @@ func (h *accessRequestHandler) Handle(action proto.Event_Type, meta *proto.Event
 		return nil
 	}
 
-	if ar.Status.Level != statusPending {
+	if ar.Status.Level != statusPending && !(ar.Status.Level == statusSuccess && ar.Metadata.State == v1.ResourceDeleting) {
```
tsjohns9, line 61: I was thinking about adding a func to figure out when to process based on status. The addition of the state checks makes me think it's time for that. I can do it though once this is merged if you want.

pkg/apic/apiserver/models/api/v1/resourceinstance.go
```diff
@@ -118,3 +118,17 @@ func (ri *ResourceInstance) FromInstance(from *ResourceInstance) error {
 func (ri *ResourceInstance) GetRawResource() json.RawMessage {
 	return ri.rawResource
 }
+
+// GetRawResource gets the resource as bytes
+func (ri *ResourceInstance) GetAPIPath() string {
```
tsjohns9, line 123: Should we also check for the selfLink, or do you think it's unnecessary? Maybe call this function `GetSelfLink`

tsjohns9 (approved):

pkg/apic/apiserver/models/api/v1/resourceinstance.go
jcollins-axway: Good call, I will actually at it to ResourceMeta, so no need to convert to an instance

pkg/apic/apiserver/models/api/v1/resourcemeta.go
```diff
@@ -57,6 +66,31 @@ func (rm *ResourceMeta) GetMetadata() Metadata {
 	return rm.Metadata
 }
 
+// GetSelfLink gets the resource metadata selflink
+func (rm *ResourceMeta) GetSelfLink() string {
+	if rm != nil {
```
tsjohns9, line 71: If rm is nil it will skip trying to find the SelfLink, and then go to the next if which checks `rm.Group`, and we will hit a nil pointer. If you want to check for nil It should return early and skip the rest of the func.

jcollins-axway: Thanks

tsjohns9 (approved):

tsjohns9 (approved):

sbolosan (approved):

vivekschauhan (approved):

PR #353, opened 2022-03-15
APIGOV-20810 - provisioning mocks

jcollins-axway:


sbolosan (approved):

pkg/util/subresource.go
```diff
@@ -33,6 +33,24 @@ func GetAgentDetails(h handler) map[string]interface{} {
...
+// if a value can not be parsed to a string its value is ignored
+func GetAgentDetailStrings(h handler) map[string]string {
+	details := GetAgentDetails(h)
+	if details == nil {
+		return nil
+	}
+
+	strMap := make(map[string]string)
+
+	for k, v := range details {
```
tsjohns9, line 46: What if something is an int or bool. Should we try to convert that to a string before skipping it?

pkg/util/subresource_test.go
```diff
@@ -42,6 +42,51 @@ func TestGetAgentDetails(t *testing.T) {
...
+		ri       *mockRI
+		expected map[string]string
+	}{
+		{
+			name:     "should return nil if no agent details are found",
+			ri:       &mockRI{subResources: map[string]interface{}{}},
+			expected: nil,
+		},
+		{
+			name: "should return nil the agent-details key is found, but is not a map[string]interface{}",
```
tsjohns9, line 57: should return nil when the ...

tsjohns9 (approved):

pkg/util/subresource.go
jcollins-axway: updated to use fmt.Sprint for all values

sbolosan (approved):

pkg/api/mockhttpclient.go
```diff
@@ -98,8 +99,9 @@ func (c *MockHTTPClient) sendMultiple(request Request) (*Response, error) {
 
 	fileName := c.Responses[c.RespCount].FileName
 
+	dat := []byte(c.Responses[c.RespCount].RespData)
```
tsjohns9, line 102: Are you adding an option for a string that you want returned for a call?

jcollins-axway: Yes, however a RespFile setting would still override this

vivekschauhan (approved):

PR #357, opened 2022-03-16
APIGOV-22058 - ARD and CRD caching

jcollins-axway:
- ARD watch handler and caching
- CRD watch handler and caching
- Fix kind/self link for unscoped resources
- Update template for watch topic resources


pkg/agent/cache/apiservice.go
```diff
@@ -0,0 +1,137 @@
...
+
+// AddAPIService - add/update APIService resource in cache
+func (c *cacheManager) AddAPIService(svc *v1.ResourceInstance) error {
+	apiID, err := util.GetAgentDetailsValue(svc, defs.AttrExternalAPIID)
+	if err != nil {
+		return fmt.Errorf("failed to get external API ID from APIService resource: %s", err)
+	}
+	if apiID != "" {
+		defer c.setCacheUpdated(true)
+		apiName, _ := util.GetAgentDetailsValue(svc, defs.AttrExternalAPIName)
```
sbolosan, line 24: you always guaranteed an apiName?  probably?

pkg/agent/handler/credentialrequestdefinition.go
```diff
@@ -0,0 +1,36 @@
...
+
+// NewCRDHandler creates a Handler for Credential Request Definitions
+func NewCRDHandler(agentCacheManager agentcache.Manager) Handler {
+	return &crdHandler{
+		agentCacheManager: agentCacheManager,
+	}
+}
+
+// Handle processes grpc events triggered for Credentials
+func (h *crdHandler) Handle(action proto.Event_Type, meta *proto.EventMeta, resource *v1.ResourceInstance) error {
```
sbolosan, line 22: this function is playing tricks on me... 
line 23, will return nil, so why check it?
line 31, if action is created, or updated, why even check for deleted?
line 32, can that return actually be an error?  (didn't check function DeleteCredentialRequestDefinitionByName

pkg/agent/cache/apiservice.go
jcollins-axway: not sure, not my code, just split it to multiple files

sbolosan: Ok..  just wondering in case since we aren't check for "" or nil.

pkg/agent/handler/credentialrequestdefinition.go
jcollins-axway: line 23: This handler is only for credential request definitions, but all handlers get invoked when an event is received.
line 31: you are right, I just mimicked other handlers 
line 32: it can be if there was an issue removing from cache

pkg/agent/stream/watchtopic.go
```diff
@@ -92,248 +132,72 @@ func getCachedWatchTopic(c cache.GetItem, key string) (*mv1.WatchTopic, error) {
...
-		],
-		"description": "Watch Topic used by a discovery agent for resources in the {{.Scope}} environment."
+func NewDiscoveryWatchTopic(name, scope, agentResourceKind string) WatchTopicValues {
+	return WatchTopicValues{
+		Name:        name,
+		Title:       name,
+		Description: fmt.Sprintf(desc, "discovery", scope),
+		Kinds: []kindValues{
+			{KindName: agentResourceKind, ScopeName: scope, EventTypes: updated},
+			{KindName: "Category", Group: "category", EventTypes: all},
```
tsjohns9, line 160: Can you please use the actual kind name, like `KindName: v1.AccessRequestGVK().Kind`

tsjohns9 (approved):

pkg/agent/cache/credentialrequestdefinitions.go
```diff
@@ -0,0 +1,33 @@
...
+
+import v1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
+
+// Credential request definition cache management
+
+// AddCredentialRequestDefinition -  add/update CredentialRequestDefinition resource in cache
+func (c *cacheManager) AddCredentialRequestDefinition(resource *v1.ResourceInstance) {
+	defer c.setCacheUpdated(true)
+
+	c.crdMap.Set(resource.Name, resource)
```
vivekschauhan, line 11: To make caches consistent can we rather use Metadata.ID as the key and use name as secondary key?. Also add methods Get method(default to return by ID)?

pkg/agent/cache/credentialrequestdefinitions.go
```diff
@@ -0,0 +1,33 @@
...
+	if item != nil {
+		if ard, ok := item.(*v1.ResourceInstance); ok {
+			return ard, nil
+		}
+	}
+	return nil, err
+}
+
+// DeleteCredentialRequestDefinitionByName - deletes the CredentialRequestDefinition cache based on resource name
+func (c *cacheManager) DeleteCredentialRequestDefinitionByName(name string) error {
```
vivekschauhan, line 29: If using Metadata.ID as key we can keep Delete method(default to use ID) and if needed (don't think for this case) we can have Delete by name

pkg/agent/cache/manager.go
```diff
@@ -61,6 +57,16 @@ type Manager interface {
 	SetAccessControlList(acl *v1.ResourceInstance)
 	GetAccessControlList() *v1.ResourceInstance
 
+	// AccessRequestDefintion cache related methods
```
vivekschauhan, line 60: I see handlers populating the cache for these new cached resources, however on initialization the discovery cache job also needs to populate them even in gRPC mode (in case the cache persistent is not enabled)

pkg/agent/cache/category.go
```diff
@@ -0,0 +1,69 @@
...
+
+	v1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
+	"github.com/Axway/agent-sdk/pkg/cache"
+)
+
+// Category cache management
+
+// AddCategory - add/update Category resource in cache
+func (c *cacheManager) AddCategory(resource *v1.ResourceInstance) {
+	c.ApplyResourceReadLock()
```
vivekschauhan, line 14: We should remove the apply and release read lock from add and delete methods, I think this was a left over mistakenly by me .

pkg/agent/stream/watchtopic.go
```diff
@@ -92,248 +132,72 @@ func getCachedWatchTopic(c cache.GetItem, key string) (*mv1.WatchTopic, error) {
...
+	return WatchTopicValues{
+		Name:        name,
+		Title:       name,
+		Description: fmt.Sprintf(desc, "traceability", scope),
+		Kinds: []kindValues{
+			{KindName: agentResourceKind, ScopeName: scope, EventTypes: updated},
+			{KindName: "APIService", ScopeName: scope, EventTypes: all},
+			{KindName: "APIServiceInstance", ScopeName: scope, EventTypes: all},
+			{KindName: "AccessRequest", ScopeName: scope, EventTypes: all},
+			{KindName: "ManagedApplication", ScopeName: scope, EventTypes: createdOrUpdated},
```
vivekschauhan, line 183: For trace agent we need the delete as well

vivekschauhan (approved):

tsjohns9 (approved):

sbolosan (approved):

PR #359, opened 2022-03-17
APIGOV-22080 Update resource handling in deleting state

tsjohns9:
Update the handlers to not fetch a managed app when in deleting state


pkg/agent/handler/accessrequest.go
```diff
@@ -60,10 +60,18 @@ func (h *accessRequestHandler) Handle(action proto.Event_Type, meta *proto.Event
 		return nil
 	}
 
-	if ok := shouldProcess(ar.Status.Level, ar.Metadata.State); !ok {
-		return nil
+	if ok := shouldProcessPending(ar.Status.Level, ar.Metadata.State); ok {
+		return h.onPending(ar)
+	}
+
+	if ok := shouldProcessDeleting(ar.Status.Level, ar.Metadata.State, len(ar.Finalizers)); ok {
```
tsjohns9, line 67: When a resource is in deleting state we end up creating another update event to remove the finalizer. So, I added a check to see if there are finalizers before deprovisioning.

sbolosan:
@tsjohns9 @jcollins-axway , can we throw in some logging for the shouldNNNN funcs?

pkg/agent/handler/accessrequest.go
```diff
@@ -74,43 +82,39 @@ func (h *accessRequestHandler) Handle(action proto.Event_Type, meta *proto.Event
 		return err
 	}
 
-	if action == proto.Event_DELETED {
```
tsjohns9, line 77: Do we need to do something on delete?

vivekschauhan (approved):

sbolosan:
In the handlers... just maybe some trace logging. There was one section where it was calling funcs should"NNNN", and would be nice to know since it was being called back to back to back

pkg/agent/handler/accessrequest.go
```diff
@@ -60,10 +58,20 @@ func (h *accessRequestHandler) Handle(action proto.Event_Type, meta *proto.Event
 		return nil
 	}
 
-	if ok := shouldProcess(ar.Status.Level, ar.Metadata.State); !ok {
-		return nil
+	if ok := shouldProcessPending(ar.Status.Level, ar.Metadata.State); ok {
+		log.Tracef("access request handler - processing resource in pending status")
```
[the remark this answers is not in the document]
sbolosan: 🤙

PR #360, opened 2022-03-20
APIGOV-20810 - remove max app creds update ard create

jcollins-axway:


sbolosan (approved):

vivekschauhan (commented):
I guess the target branch should be APIGOV-20810 instead of main

jcollins-axway:
> I guess the target branch should be APIGOV-20810 instead of main

Yes, thanks

tsjohns9 (approved):

pkg/apic/provisioning/accessrequestdefinitionbuilder.go
```diff
@@ -35,10 +37,20 @@ func NewAccessRequestBuilder(registerFunc RegisterAccessRequestDefinition) Acces
 
 // SetName - set the name of the access request
 func (a *accessRequestDef) SetName(name string) AccessRequestBuilder {
+	if a.title == "" {
```
vivekschauhan, line 40: Do we need this check/assignment here as the Register method is already doing this anyways?

vivekschauhan (approved):

PR #363, opened 2022-03-22
APIGOV-22093 Update GRPC Handlers and tests

tsjohns9:
- Set status to error on resources when processing fails
- Update tests to handle deleting state


pkg/agent/handler/accessrequest.go
```diff
@@ -88,41 +88,50 @@ func (h *accessRequestHandler) onPending(ar *mv1.AccessRequest) error {
...
+		return
 	}
+
 	status := h.prov.AccessRequestDeprovision(req)
 
 	ri, _ := ar.AsInstance()
 	if status.GetStatus() == prov.Success {
 		h.client.UpdateResourceFinalizer(ri, arFinalizer, "", false)
+	} else {
+		h.onError(ar, fmt.Errorf(status.GetMessage()))
```
tsjohns9, line 127: Set status to error when deprovision failed

pkg/util/subresource.go
```diff
@@ -61,6 +62,11 @@ func GetAgentDetailsValue(h handler, key string) (string, error) {
 // Returns errors if unable to perform type conversion.
 // Returns an empty string if the value does not exist, or if there is an error.
 func GetSubResourcePropertyValue(h handler, subRes, key string) (string, error) {
+	// check for a nil value, or a pointer to a nil value
+	if h == nil || reflect.ValueOf(h).Kind() == reflect.Ptr && reflect.ValueOf(h).IsNil() {
```
tsjohns9, line 66: Checking for nil, or a nil pointer to avoid panics

pkg/util/util.go
```diff
@@ -248,3 +250,12 @@ func ToString(v interface{}) string {
 	}
 	return s
 }
+
+// AssertError asserts on the existence of an error for tests
+func AssertError(t *testing.T, hasErr bool, err error) {
```
jcollins-axway, line 255: I don't like pulling in the testing and assert packages for releases, that means all agents will have this package as well. Could we revert the use of this function?

jcollins-axway (changes requested):

pkg/util/subresource.go
jcollins-axway: maybe this check is worth moving to util function?

tsjohns9: I like that idea

pkg/util/util.go
tsjohns9: Yeah

pkg/util/subresource.go
```diff
@@ -63,7 +62,7 @@ func GetAgentDetailsValue(h handler, key string) (string, error) {
 // Returns an empty string if the value does not exist, or if there is an error.
 func GetSubResourcePropertyValue(h handler, subRes, key string) (string, error) {
 	// check for a nil value, or a pointer to a nil value
-	if h == nil || reflect.ValueOf(h).Kind() == reflect.Ptr && reflect.ValueOf(h).IsNil() {
+	if isNil := IsNil(h); isNil {
```
jcollins-axway, line 65: if the func returns a bool, no need for this var.

How are you pulling this in from the util package?

jcollins-axway: This is the util package, haha

jcollins-axway (approved):

pkg/util/subresource.go
tsjohns9: haha :)
I'll remove the var

PR #367, opened 2022-03-24
APIGOV-02468 - validate subscription state description

sbolosan:
1. validate length of description > 350
2. if greater, warn that we are truncating
3. send in truncated description


tsjohns9 (approved):

tsjohns9:
Maybe a test for it too since it seems this might have been here before but removed.

sbolosan:
@tsjohns9 , can do... i went back to its inception and unfortunately there wasn't a check... but will add a test

jcollins-axway (approved):

PR #369, opened 2022-03-25
APIGOV-22160 -Cache should store items that are immutable

sbolosan:


vivekschauhan (approved):

pkg/cache/cache.go
```diff
@@ -307,12 +323,17 @@ func (c *itemCache) set(thisAction cacheAction) (thisReply cacheReply) {
...
 	}
+
 	c.Items[key] = &Item{
 		Object:        data,
 		UpdateTime:    time.Now().Unix(),
 		Hash:          hash,
 		SecondaryKeys: secKeys,
 	}
+
+	if data != nil && reflect.ValueOf(data).Type().Kind() == reflect.Ptr {
```
tsjohns9, line 334: You might need to also check if you have a pointer to nil. I added a func for it in the marketplace branch in util/util.go. It checks for nil, and a pointer that has a nil value. Turns out these are two different things.

```
// IsNil checks a value, or a pointer for nil
func IsNil(v interface{}) bool {
	return v == nil || reflect.ValueOf(v).Kind() == reflect.Ptr && reflect.ValueOf(v).IsNil()
}
```

tsjohns9 (approved):

PR #370, opened 2022-03-28
APIGOV-20810 - add interface methods for request data

jcollins-axway:
- Access Request data, which matches the schema in the access request definition
- Credential data, which matches the request schema in the credential request definition
- Public key oauth support

### DIFF
--- a/pkg/agent/provisioning.go
+++ b/pkg/agent/provisioning.go
@@ -17,6 +17,9 @@ const (
 
 // createOrUpdateCredentialRequestDefinition -
 func createOrUpdateCredentialRequestDefinition(data *v1alpha1.CredentialRequestDefinition) (*v1alpha1.CredentialRequestDefinition, error) {
+	if !agent.agentFeaturesCfg.MarketplaceProvisioningEnabled() {
+		return nil, nil
+	}
 	crdRI, _ := agent.cacheManager.GetCredentialRequestDefinitionByName(data.Name)
 	if crdRI == nil {
 		return agent.apicClient.RegisterCredentialRequestDefinition(data, false)
@@ -77,6 +80,9 @@ func NewOAuthCredentialRequestBuilder() provisioning.CredentialRequestBuilder {
 
 // createOrUpdateAccessRequestDefinition -
 func createOrUpdateAccessRequestDefinition(data *v1alpha1.AccessRequestDefinition) (*v1alpha1.AccessRequestDefinition, error) {
+	if !agent.agentFeaturesCfg.MarketplaceProvisioningEnabled() {
+		return nil, nil
+	}
 	ardRI, _ := agent.cacheManager.GetAccessRequestDefinitionByName(data.Name)
 	if ardRI == nil {
 		return agent.apicClient.RegisterAccessRequestDefinition(data, false)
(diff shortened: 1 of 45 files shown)


sbolosan (approved):

vivekschauhan (approved):

pkg/apic/provisioning/credentialrequest.go
```diff
@@ -10,4 +10,6 @@ type CredentialRequest interface {
 	GetCredentialDetailsValue(key string) string
 	// GetCredentialType returns the type of credential related to this request
 	GetCredentialType() string
+	// GetCredentialData returns the map[string]interface{} of data from the request
+	GetCredentialData() map[string]interface{}
```
jcollins-axway, line 14: @tsjohns9 this change ant he GetAccessRequestData above are meant to return the data that matches the request schema in the *RequestDefinition resources

pkg/agent/provisioning.go
```diff
@@ -34,46 +28,118 @@ func createOrUpdateCredentialRequestDefinition(data *v1alpha1.CredentialRequestD
...
+type crdBuilderOptions struct {
+	name      string
+	provProps []provisioning.PropertyBuilder
+	reqProps  []provisioning.PropertyBuilder
+}
+
 // NewCredentialRequestBuilder - called by the agents to build and register a new credential reqest definition
-func NewCredentialRequestBuilder() provisioning.CredentialRequestBuilder {
-	return provisioning.NewCRDBuilder(createOrUpdateCredentialRequestDefinition)
+func NewCredentialRequestBuilder(options ...func(*crdBuilderOptions)) provisioning.CredentialRequestBuilder {
```
tsjohns9, line 38: A specific type might be helpful for these funcs

pkg/agent/provisioning.go
```diff
@@ -34,46 +28,118 @@ func createOrUpdateCredentialRequestDefinition(data *v1alpha1.CredentialRequestD
...
-		SetName(apikeyCRD).
-		SetProvisionSchema(provisioning.NewSchemaBuilder().
-			AddProperty(
-				provisioning.NewSchemaPropertyBuilder().
-					SetName("key").
-					SetLabel("API Key").
-					SetRequired().
-					IsString().
-					IsEncrypted()))
+func NewAPIKeyCredentialRequestBuilder(options ...func(*crdBuilderOptions)) provisioning.CredentialRequestBuilder {
```
tsjohns9, line 112: Do agents need to be updated?

tsjohns9 (approved):

pkg/apic/provisioning/definitions.go
```diff
@@ -1,5 +1,19 @@
...
+	APIKeyARD         = "api-key"
+	APIKeyCRD         = "api-key"
+	OAuthSecretCRD    = "oauth-secret"
+	OAuthPublicKeyCRD = "oauth-public-key"
+
+	OauthClientID     = "clientId"
+	OauthClientSecret = "clientSecret"
+	OauthPublicKey    = "publicKey"
+
+	APIKey = "apiKey"
```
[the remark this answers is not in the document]
jcollins-axway: The "api-key" here is the name of the resources, as we tend to name resources separated by dashes. The "apiKey" is the field name in the request definition as we tend to name fields in camel case.

pkg/agent/provisioning.go
jcollins-axway: Unless you need to add options, not for this call.

pkg/agent/provisioning.go
tsjohns9: Actually, I'm not sure how an agent can implement this func because the param is a private type

jcollins-axway: They use the With functions that return that function. They do not need to know that type.

tsjohns9: I see a few exported funcs like `WithCRDProvisionSchemaProperty ` that return `func(c *crdBuilderOptions)` that an agent can use. But if an agent needs their own option func then they won't be able to implement it. Not sure if they need it or not, just bringing it up.

jcollins-axway: Nah, I think we will be the only ones providing these options

tsjohns9 (approved):

PR #386, opened 2022-04-08
APIGOV-00002 - disable commit/push of version file for now

dfeldick:
Until we can push directly to main, I commented out the commit/push from the script. In it's place, I added echo statements to instruct the user exactly what to manually do to complete the release.

### DIFF
--- a/build/scripts/promote.sh
+++ b/build/scripts/promote.sh
@@ -44,11 +44,21 @@ update_version_file() {
 }
 
 commit_promotion() {
-    echo "Committing the new promoted version to main"
-    cd ${BASE_DIR}
-    git add version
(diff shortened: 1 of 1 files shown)


build/scripts/promote.sh
```diff
@@ -44,11 +44,21 @@ update_version_file() {
 }
```
jcollins-axway, line 44: Seems that there are github actions that can create a Branch, commit, and stat a PR.

jcollins-axway (approved):

build/scripts/promote.sh
dfeldick: can probably do that in my script to. Just don't have time to do it right now.

PR #387, opened 2022-04-08
New logger interface and implementation

tsjohns9:
Created a new logging interface and struct that acts like the Logrus FieldLogger, and handles logp writing like we do currently so that additional fields can be added to log messages


pkg/util/log/logger.go
```diff
@@ -0,0 +1,344 @@
...
+// WithError adds an error field to the message
+func (l *logger) WithError(err error) FieldLogger {
+	return &logger{entry: l.entry.WithError(err)}
+}
+
+// Debugf prints a formatted debug message
+func (l *logger) Debugf(format string, args ...interface{}) {
+	if l.isLogP() {
+		args = append(args, l.formatEntries()...)
+		logp.L().Debugf(format, args)
```
tsjohns9, line 78: @jcollins-axway The original log package uses a const called `debugSelector` when writing debug logs to logp. I don't know exactly what we are trying to do with that. Do you mind updating this PR with anything else we may need to keep it consistent with the current implementation?

pkg/util/log/logger.go
```diff
@@ -0,0 +1,344 @@
...
+	Infof(format string, args ...interface{})
+	Infoln(args ...interface{})
+
+	Warn(args ...interface{})
+	Warnf(format string, args ...interface{})
+	Warnln(args ...interface{})
+}
+
+// Redactor interface for redacting log messages
+type Redactor interface {
```
tsjohns9, line 40: Should the `Redactor` be added to the `FieldLogger` interface?

pkg/util/log/logger.go
jcollins-axway: Logp does not have trace, so to differ the 2 logging events the debugSelector is used to name the logger in logp.

pkg/util/log/logger.go
jcollins-axway: The AMPGW team added these, would want to check to see how they are using this.

jcollins-axway (dismissed):

tsjohns9:
I updated a few components to use the FieldLogger, like the jobs, stream, cache manager, and client

jcollins-axway (dismissed):

sbolosan (dismissed):

sbolosan (approved):

pkg/transaction/metric/metricscollector.go
```diff
@@ -161,8 +166,17 @@ func (c *collector) Execute() error {
 	c.usageEndTime = now()
 	c.metricEndTime = now()
 	c.orgGUID = c.getOrgGUID()
-	log.Debugf("Generating usage event [start timestamp: %d, end timestamp: %d]", util.ConvertTimeToMillis(c.usageStartTime), util.ConvertTimeToMillis(c.usageEndTime))
-	log.Debugf("Generating metric event [start timestamp: %d, end timestamp: %d]", util.ConvertTimeToMillis(c.metricStartTime), util.ConvertTimeToMillis(c.metricEndTime))
+	c.logger.
+		WithField("start timestamp", util.ConvertTimeToMillis(c.usageStartTime)).
```
jcollins-axway, line 170: I know we have this in other places, but wondering if we should skip the spaces for the field names.  I know json will still be valid, just thinking we may want to make it a standard?

jcollins-axway (approved):

PR #389, opened 2022-04-11
INT - new logger and more logging in stream handlers

jcollins-axway:


pkg/agent/handler/proxy.go
```diff
@@ -36,14 +40,47 @@ func (h *StreamWatchProxyHandler) UnregisterTargetHandler(name string) {
...
 			if err != nil {
 				return err
 			}
 		}
 	}
 	return nil
 }
+
+// NewEventContext - create a context for the new event
+func NewEventContext(action proto.Event_Type, eventMetadata *proto.EventMeta, kind, name string) context.Context {
```
tsjohns9, line 56: Can these go into handler.go

tsjohns9 (approved):

pkg/agent/stream/eventlistener.go
```diff
@@ -97,19 +97,16 @@ func (em *EventListener) start() (done bool, err error) {
 
 // handleEvent fetches the api server ResourceClient based on the event self link, and then tries to save it to the cache.
 func (em *EventListener) handleEvent(event *proto.Event) error {
-	em.logger.
-		WithField("sequenceID", event.Metadata.SequenceID).
-		WithField("action", proto.Event_Type_name[int32(event.Type)]).
-		WithField("kind", event.Payload.Kind).
-		WithField("name", event.Payload.Name).
-		Debug("processing received watch event")
+	ctx := handler.NewEventContext(proto.Event_Type(event.Type), event.Metadata, event.Payload.Name, event.Payload.Kind)
```
tsjohns9, line 100: `event.Type` is already a `proto.Event_Type` so this conversion is unnecessary

jcollins-axway: the casting from the previous log had me thinking it was an int32, you  right though. Fixed

PR #391, opened 2022-04-12
APIGOV-21522 - Lookup for subscription and consumer orgID

vivekschauhan:
- Changes to use references sub resource in AccessRequest to lookup associated subscription 
- Changes to use marketplace sub resource in ManagedApplication to lookup consumer org identifier

### DIFF
--- a/pkg/agent/discoverycache.go
+++ b/pkg/agent/discoverycache.go
@@ -8,6 +8,7 @@ import (
 	"github.com/Axway/agent-sdk/pkg/migrate"
 	"github.com/Axway/agent-sdk/pkg/util"
 
+	"github.com/Axway/agent-sdk/pkg/apic/definitions"
 	defs "github.com/Axway/agent-sdk/pkg/apic/definitions"
 
 	"github.com/Axway/agent-sdk/pkg/agent/resource"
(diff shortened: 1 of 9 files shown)


sbolosan (dismissed):

jcollins-axway (dismissed):

pkg/agent/discoverycache.go
```diff
@@ -344,7 +344,7 @@ func (j *discoveryCache) updateManagedApplicationCache() {
 	// TODO - Remove custom subresource and include subject subresource when added to model
```
jcollins-axway, line 344: These comments still needed?

pkg/agent/handler/traceaccessrequest_test.go
```diff
@@ -94,9 +95,9 @@ func TestTraceAccessRequestTraceHandler(t *testing.T) {
 	c.getRI = &v1.ResourceInstance{
 		ResourceMeta: v1.ResourceMeta{
 			Metadata: v1.Metadata{
-				ID: "subscription",
+				ID: "subscription-id",
```
jcollins-axway, line 98: Should these be constants?

jcollins-axway (approved):

pkg/agent/handler/traceaccessrequest_test.go
vivekschauhan: Its test data, but have other string that I can create constants for. Will do the change

pkg/agent/discoverycache.go
vivekschauhan: Will remove it

sbolosan (approved):

PR #399, opened 2022-04-26
APIGOV-22452 - handle nil pointer when central config hasn't been created

sbolosan:
Nil check when central config hasn't been created/initialized yet.  As in when EXE subcommands are ran.


sbolosan:
Check for nil.  There is a case where init is being called before the cmd initialization. When this happens the singleton object will not be created. ie., .exe subcommands. Just return nil. The next call to GetMetricCollector should create the singleton object

dfeldick (approved):

vivekschauhan (approved):

PR #402, opened 2022-05-04
INT - change how things are logged when skipping a UC sub

jcollins-axway:


sbolosan (approved):
dave will be glad this is updated.

PR #404, opened 2022-05-05
APIGOV-22377 - updates for handling marketplace provisioning error paths

jcollins-axway:
- raise deprecation warning if DOSA_ is used
- throw config error if DOSA is used with Marketplace Provisioning
- Return error status for Access Request and Credential when linked Managed App not Success

### DIFF
--- a/pkg/agent/handler/credential_test.go
+++ b/pkg/agent/handler/credential_test.go
@@ -38,6 +38,7 @@ func TestCredentialHandler(t *testing.T) {
 		name             string
 		outboundStatus   string
 		subError         error
+		appStatus        string
 	}{
 		{
 			action:           proto.Event_CREATED,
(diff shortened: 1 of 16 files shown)


sbolosan (approved):

pkg/agent/handler/credential_test.go
```diff
@@ -98,6 +106,11 @@ func TestCredentialHandler(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			credApp.SubResources["status"].(map[string]interface{})["level"] = prov.Success.String()
+			if tc.appStatus != "" {
```
tsjohns9, line 110: What are these lines for?

tsjohns9 (approved):

pkg/agent/handler/credential_test.go
jcollins-axway: Setting the status to success if an appStatus is not specified for the test, otherwise use the appStatus from the test case

PR #414, opened 2022-05-16
APIGOV-22588 clear cache and re-run cache discovery when harvester returns an error

tsjohns9:
- discoveryCache job is now a channel job that gets executed when a signal is received
- poll client and stream clients trigger trigger the discoveryCache job when harvester is down, and trigger a re-sync of the cache


pkg/agent/cache/manager.go
```diff
@@ -103,23 +104,23 @@ type Manager interface {
...
 	hasLoadedPersistedCache bool
+	instanceMap             cache.Cache
 	isCacheUpdated          bool
+	logger                  log.FieldLogger
+	managedApplicationMap   cache.Cache
+	persistedCache          cache.Cache
+	resourceCacheReadLock   sync.Mutex
+	sequenceCache           cache.Cache
+	subscriptionMap         cache.Cache
+	teams                   cache.Cache
```
jcollins-axway, line 123: Not sure if this is something goland did for you or not, but I once read an article that talked about how golang reserves memory for attributes in a struct.  

It does not matter for this, but something interesting to keep in mind in the future. 

https://medium.com/techverito/golang-struct-size-and-memory-optimisation-b46b124f008d

jcollins-axway (dismissed):

vivekschauhan (approved):

jcollins-axway (approved):

pkg/agent/discoverycache.go
```diff
@@ -85,8 +93,22 @@ func (j *discoveryCache) Status() error {
 	return fmt.Errorf("could not establish a connection to APIC to update the cache")
 }
 
-// Execute -
+// Execute - starts a loop to wait for the discovery cache to receive a signal to rebuild the cache
 func (j *discoveryCache) Execute() error {
+	for {
```
vivekschauhan, line 98: For loop is not required here since select will be blocking in this case, so it can be removed

tsjohns9: I think the for is needed so that it can come back to waiting for a signal. Without the for loop it will wait here for one write from either stop or signal, and after that execute will end

PR #415, opened 2022-05-16
APIGOV- 22585 - marketplace migration

sbolosan:


sbolosan:
1.  Need to fix current test failures
2. Need to create marketplace migration tests

sbolosan:
I had to change a handful of spec stuff to be public so that the migration could access it... let me know if there is a better way to get at the funcs

pkg/agent/discoverycache.go
```diff
@@ -125,10 +127,20 @@ func (j *discoveryCache) updateAPICache() error {
...
+	// Check if marketplace provisioning is enabled.  If it is enabled, verify that current apiservices have the necessary request definitions
+	if agent.agentFeaturesCfg.MarketplaceProvisioningEnabled() {
+		j.logger.Trace("marketplace provisioning migration starting")
+		j.marketplaceProvisioningMigration(apiServices)
+	}
+
+	return j.attributeMigration(apiServices, existingAPIs)
+}
+
+func (j *discoveryCache) attributeMigration(apiServices []*apiV1.ResourceInstance, existingAPIs map[string]bool) error {
```
tsjohns9, line 139: Maybe we can have an array of migrate types that implement a `Migration` interface that we can loop through and pass each service too so we can keep the specifics of what is being migrated out of the discovery cache

pkg/migrate/marketplacemigration.go
```diff
@@ -0,0 +1,367 @@
...
+	"github.com/Axway/agent-sdk/pkg/apic"
+	v1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
+	mv1a "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/management/v1alpha1"
+	"github.com/Axway/agent-sdk/pkg/apic/provisioning"
+	"github.com/Axway/agent-sdk/pkg/config"
+	"github.com/Axway/agent-sdk/pkg/util/log"
+)
+
+// MarketplaceMigrator interface for performing an marketplace provisioning migration
+type MarketplaceMigrator interface {
```
tsjohns9, line 20: Can we have one interface that both migration types interface

pkg/agent/discoverycache.go
tsjohns9: Or we can add another type that implements the `Migration` interface that contains a list of all the migrations to run, and then as we loop through each service we just pass the service into `Migrate` one time, and that type will handle all the other migrations.

pkg/migrate/marketplacemigration.go
```diff
@@ -0,0 +1,367 @@
...
+
+			err := fun(ri)
+			errCh <- err
+		}(f)
+	}
+
+	wg.Wait()
+	close(errCh)
+
+	for e := range errCh {
```
tsjohns9, line 70: Looping through the funcs and the error channel might not be needed since you only have one func in your funcs array

pkg/migrate/marketplacemigration.go
```diff
@@ -0,0 +1,367 @@
...
+		wg.Add(1)
+
+		go func(ri *v1.ResourceInstance) {
+			defer wg.Done()
+
+			apiSvcInst := mv1a.NewAPIServiceInstance(ri.Name, ri.Metadata.Scope.Name)
+			apiSvcInst.FromInstance(ri)
+
+			specDefintionType := resourceInstance.Spec["definition"].(map[string]interface{})["type"].(string)
+			specDefinitionValue := resourceInstance.Spec["definition"].(map[string]interface{})["value"].(string)
```
tsjohns9, line 142: You might want to check if the type conversion was ok for both of these

pkg/migrate/marketplacemigration.go
```diff
@@ -0,0 +1,367 @@
...
+		return err
+	}
+
+	wg := &sync.WaitGroup{}
+	errCh := make(chan error, len(resources))
+
+	for _, resource := range resources {
+		wg.Add(1)
+
+		go func(ri *v1.ResourceInstance) {
```
tsjohns9, line 135: It might be a good idea to break this func down a little bit into smaller funcs since it is kinda big.

pkg/migrate/marketplacemigration.go
sbolosan: You right... i had multiple funcs, but not anymore

pkg/migrate/marketplacemigration.go
sbolosan: Yup... in the process... i agree.

pkg/migrate/marketplacemigration.go
sbolosan: Done...

sbolosan:
@tsjohns9 @dfeldick @jcollins-axway @vivekschauhan moved from draft to review...  can you guys look over this please.  thanks in advance

tsjohns9 (approved):

pkg/agent/provisioning.go
```diff
@@ -1,24 +1,77 @@
...
 // credential request definitions
 // createOrUpdateDefinition -
-func createOrUpdateDefinition(data v1.Interface) (*v1.ResourceInstance, error) {
+func createOrUpdateDefinition(data v1.Interface, marketplaceMigration migrate.Migrator) (*v1.ResourceInstance, error) {
 	if agent.agentFeaturesCfg == nil || !agent.agentFeaturesCfg.MarketplaceProvisioningEnabled() {
 		return nil, nil
 	}
-	return agent.apicClient.CreateOrUpdateResource(data)
+
+	var runMarketplaceMigration = false
```
jcollins-axway, line 22: Can this migration check be moved to its own func?

Wy not checks for access request defs? What if an agent makes a blanket one like the credential request defs are?  such as api-key?

pkg/agent/provisioning.go
```diff
@@ -1,24 +1,77 @@
...
+
+	// Check (only) credential request definition to see if it exists prior to CreateOrUpdateResource call
+	ri, err := data.AsInstance()
+	if mv1a.CredentialRequestDefinitionGVK().Kind == ri.Kind {
```
jcollins-axway, line 26: if in fact we do only care about market place migration then this can be moved into the createOrUpdateCredentialRequestDefinition func.

pkg/agent/provisioning.go
sbolosan: It can be...
...and yes it should now check for ARDs

pkg/agent/provisioning.go
sbolosan: Good point.  Will do

pkg/apic/specparser.go
```diff
@@ -13,29 +13,34 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-type specProcessor interface {
-	getEndpoints() ([]EndpointDefinition, error)
+// SpecProcessor -
+type SpecProcessor interface {
```
jcollins-axway, line 17: Maybe not now, but considering the spec parser is not just used by apic, maybe we move it to its own package, util/specparser/*.go?

pkg/migrate/marketplacemigration.go
```diff
@@ -0,0 +1,424 @@
...
+	apiSvcInst := mv1a.NewAPIServiceInstance(svcInstance.Name, svcInstance.Metadata.Scope.Name)
+	apiSvcInst.FromInstance(svcInstance)
+
+	specProcessor, err := m.getSpecParser(revision)
+	if err != nil {
+		return err
+	}
+
+	endPoints, err := specProcessor.GetEndpoints()
+	instanceSpecEndPoints, err := m.createInstanceEndpoint(endPoints)
```
jcollins-axway, line 286: Why do we need the endpoint checks? This is only for u[dating ARDs and CRDs within an instance, does not seem necessary to do this.

pkg/migrate/marketplacemigration.go
```diff
@@ -0,0 +1,424 @@
...
+
+	return endPoints, nil
+}
+
+func (m *MarketplaceMigration) handleSvcInstance(
+	svcInstance *v1.ResourceInstance, revision *v1.ResourceInstance,
+) error {
+	logger := m.logger.
+		WithField("instance-name", svcInstance.Name).
+		WithField("revision-name", revision.Name)
```
jcollins-axway, line 275: can we add the service-name here too?

pkg/apic/specparser.go
sbolosan: Yes

pkg/migrate/marketplacemigration.go
sbolosan: I'll have to recheck. If I didn't add in the endpoints when updating the spec, it puked on me.

pkg/migrate/marketplacemigration.go
sbolosan: Will do 👍🏾

pkg/migrate/marketplacemigration.go
```diff
@@ -0,0 +1,424 @@
...
+		}
+
+		existingARD, _ := m.cache.GetAccessRequestDefinitionByName(ardRIName)
+		if existingARD != nil && apiSvcInst.Spec.AccessRequestDefinition == "" {
+			logger.Debugf("adding the following access request definition %s", ardRIName)
+			updateRequestDefinition = true
+		}
+
+		if updateRequestDefinition {
+			inInterface := m.newInstanceSpec(instanceSpecEndPoints, revision.Name, ardRIName, credentialRequestDefinitions)
```
jcollins-axway, line 344: apiSvcInst.Spec.Endpoint?

sbolosan: Crap, yup, you're right.  Will fix

jcollins-axway:
Maybe Vivek will have other feedback, but thinking during the first run through of the migration we can add x-agent-details data about the details needed to know what ard/crd would have to be added. Then we can utilize that as a signal for subsequent calls from provisioning and restarts of the agent?  Just thoughts....

pkg/agent/provisioning.go
sbolosan: Moved to its own func...

pkg/migrate/marketplacemigration.go
sbolosan: done... removed any new instance spec creation

pkg/migrate/marketplacemigration.go
sbolosan: We actually log service upstream... so I don't believe we need it here.

pkg/migrate/marketplacemigration.go
sbolosan: using current spec from apiserviceinstance

pkg/migrate/marketplacemigration.go
jcollins-axway: We do, but you can't aggregate all logs together based on service name if not all logs log it.

pkg/apic/specparser.go
sbolosan: logged ticket https://jira.axway.com/browse/APIGOV-22680 to take care of this plus context creation/use

vivekschauhan (dismissed):

jcollins-axway (approved):

PR #420, opened 2022-05-19
APIGOV-22588 refresh cache when harvester is down for stream and poll mode

tsjohns9:
- Discovery cache is no longer a job. It is a struct with an execute method that should be called as needed
- Poll mode and stream mode will invoke discoveryCache.execute whenever a failure is encountered from harvester.
- Updated methods on the ServiceClient to take a `v1.Interface` for create/update/delete of resource.


pkg/apic/client.go
```diff
@@ -87,15 +87,19 @@ type Client interface {
...
+
 	CreateSubResource(rm v1.ResourceMeta, subs map[string]interface{}) error
 	GetResource(url string) (*v1.ResourceInstance, error)
-	CreateResource(url string, bts []byte) (*v1.ResourceInstance, error)
-	UpdateResource(url string, bts []byte) (*v1.ResourceInstance, error)
 	UpdateResourceFinalizer(ri *v1.ResourceInstance, finalizer, description string, addAction bool) (*v1.ResourceInstance, error)
-	CreateOrUpdateResource(v1.Interface) (*v1.ResourceInstance, error)
 	IsMarketplaceSubsEnabled() bool
+
+	UpdateResourceInstance(ri v1.Interface) (*v1.ResourceInstance, error)
```
tsjohns9, line 96: Updated ResourceInstance methods on the client to take in a `v1.Interface`, and deprecated the `CreateResource` and `UpdateResource`

pkg/agent/discovery_test.go
```diff
@@ -18,15 +17,12 @@ import (
...
-	hc "github.com/Axway/agent-sdk/pkg/util/healthcheck"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestDiscoveryCache(t *testing.T) {
-	dcj := newDiscoveryCache(nil, true, &sync.Mutex{}, nil)
-	dcj.getHCStatus = func(_ string) hc.StatusLevel {
-		return hc.OK
-	}
+	dcj := newDiscoveryCache(nil, nil)
```
jcollins-axway, line 24: I guess the 'j' in 'dcj' no longer applies.

pkg/agent/events/watchtopic.go
```diff
@@ -195,25 +195,14 @@ func parseWatchTopicTemplate(values WatchTopicValues) (*mv1.WatchTopic, error) {
...
-
-	var ri *v1.ResourceInstance
 	if wt.Metadata.ID != "" {
-		// delete/create required for harvester
-		ri, err := wt.AsInstance()
-		if err != nil {
-			return nil, err
-		}
-		err = rc.DeleteResourceInstance(ri)
+		err := rc.DeleteResourceInstance(wt)
```
jcollins-axway, line 199: Using that interface, makes it so nice

jcollins-axway (dismissed):

vivekschauhan (approved):

jcollins-axway (dismissed):

pkg/agent/poller/client.go
```diff
@@ -23,14 +23,16 @@ type PollClient struct {
...
-	onClientStop       func()
+	onClientStop       onClientStopCb
 	onStreamConnection OnStreamConnection
 	poller             *manager
 	seq                events.SequenceProvider
 	topicSelfLink      string
 	newPollManager     newPollManagerFunc
 }
 
+type onClientStopCb func() error
```
jcollins-axway, line 34: Should this be public so the agent package may use it as well?

jcollins-axway: maybe this is not the same type, but same signature of the cacheSyncFunc

PR #421, opened 2022-05-20
[AMPGW-814] Changes for Amplify Gateway Governance Agent provisioning flows.

mcollins-axway:


pkg/config/centralconfig.go
```diff
@@ -115,6 +115,7 @@ type CentralConfig interface {
 	IsUsingGRPC() bool
 	GetGRPCHost() string
 	GetGRPCPort() int
+	IsGRPCInsecure() bool
```
mcollins-axway, line 118: We needed this to use the new central watch mock service in our tests.

pkg/agent/handler/instance.go
```diff
@@ -26,7 +26,7 @@ func (h *instanceHandler) Handle(ctx context.Context, _ *proto.EventMeta, resour
 		return nil
 	}
 
-	if action == proto.Event_CREATED || action == proto.Event_UPDATED {
+	if action == proto.Event_CREATED || action == proto.Event_UPDATED || action == proto.Event_SUBRESOURCEUPDATED {
```
mcollins-axway, line 29: The vapi-controller in Central creates the APIServiceInstance via a POST, followed quickly by a PUT to create the x-agent-details sub-resource. If we ignore the SUBRESOURCEUPDATED event it is possible that the agent SDK pulls the main resource into the cache before it has the x-agent-details subresource. Are there any impacts to adding this ? It seems to work OK.

jcollins-axway: No impacts that I can think of

jcollins-axway (approved):

PR #423, opened 2022-05-23
APIGOV-22589

tsjohns9:
- Update discovery cache to save resources based on the provided handlers.
- Fetch marketplace provisioning resources and pass them to the handlers in sequential order.
- Register handlers based on the agent type


pkg/agent/agent.go
```diff
@@ -423,19 +432,28 @@ func newHandlers() []handler.Handler {
...
 		agent.proxyResourceHandler,
 	}
 
+	if agent.cfg.GetAgentType() == config.DiscoveryAgent {
+		handlers = append(
+			handlers,
+			handler.NewCategoryHandler(agent.cacheManager),
+			handler.NewCRDHandler(agent.cacheManager),
+			handler.NewARDHandler(agent.cacheManager),
+			// handler.NewACLHandler(agent.cacheManager),
```
tsjohns9, line 445: @jcollins-axway is this still needed?

pkg/agent/agent.go
```diff
@@ -423,19 +432,28 @@ func newHandlers() []handler.Handler {
...
 		handler.NewInstanceHandler(agent.cacheManager),
-		handler.NewCategoryHandler(agent.cacheManager),
 		handler.NewAgentResourceHandler(agent.agentResourceManager),
-		handler.NewCRDHandler(agent.cacheManager),
-		handler.NewARDHandler(agent.cacheManager),
-		// handler.NewACLHandler(agent.cacheManager),
 		agent.proxyResourceHandler,
 	}
 
+	if agent.cfg.GetAgentType() == config.DiscoveryAgent {
```
tsjohns9, line 439: Only registering handlers based on what is defined in the watch topic for the agent type.

pkg/agent/agent.go
jcollins-axway: I kept it to enable again when we do fix the update handling of ACLs on API Server

sbolosan, line 445: @jcollins-axway  this is still here for future use, yeah?

sbolosan (dismissed):

pkg/agent/agent.go
jcollins-axway: yup

sbolosan (approved):

pkg/agent/discoverycache.go
```diff
@@ -24,264 +25,428 @@ const (
...
+		return nil
+	}
+
+	dc.logger.WithField("kind", "Credential").Trace("fetching Credentials and updating cache")
+
+	credLink := mv1.NewCredential("", dc.envName).GetKindLink()
+	url := dc.formatResourceURL(credLink)
+
+	credentials, err := dc.client.GetAPIV1ResourceInstancesWithPageSize(
+		nil, url, apiServerPageSize,
```
tsjohns9, line 404: @jcollins-axway Are there certain fields to query for on the credential like we do for other resources?

pkg/agent/discoverycache.go
```diff
@@ -24,264 +25,428 @@ const (
...
 
-func (j *discoveryCache) updateCategoryCache() {
-	j.logger.Trace("updating category cache")
+func (dc *discoveryCache) handleAccessControlList() error {
+	dc.logger.WithField("kind", "AccessControlList").Trace("fetching AccessControlList and updating cache")
 
 	// Update cache with published resources
-	existingCategories := make(map[string]bool)
 	query := map[string]string{
 		apic.FieldsKey: apiServerFields,
```
tsjohns9, line 258: @jcollins-axway Are these the right fields to query for on an access control list?

pkg/agent/discoverycache.go
```diff
@@ -24,264 +25,428 @@ const (
...
 	)
+	if err != nil {
+		e := utilErrors.Wrap(ErrUnableToGetAPIV1Resources, err.Error()).FormatError("APIServiceInstances")
+		dc.logger.Error(e)
+		return e
+	}
+
+	for _, instance := range serviceInstances {
+		if err := dc.handleResource(instance); err != nil {
+			return err
```
tsjohns9, line 221: @jcollins-axway for all of these funcs I am returning an error if the handler failed to process the resource. Maybe that isn't necessary and we should just log the error if that happens. I don't think everything needs to stop. What do you think?

vivekschauhan (approved):

pkg/agent/discoverycache.go
jcollins-axway: I agree, I think we need to look at these situations in the migration code as well. Was thinking about this after talking to Shane about the issue Dave hit.

pkg/agent/discoverycache.go
jcollins-axway: I do not think we should limit the fields we get.

jcollins-axway (dismissed):

pkg/agent/discoverycache.go
tsjohns9: @jcollins-axway Can we remove the query for all the resources we fetch then?

sbolosan (dismissed):

jcollins-axway (approved):

sbolosan (approved):

pkg/agent/events/watchtopic.go
```diff
@@ -65,8 +65,8 @@ var (
 
 // getOrCreateWatchTopic attempts to retrieve a watch topic from central, or create one if it does not exist.
 func getOrCreateWatchTopic(name, scope string, client APIClient, features watchTopicFeatures) (*mv1.WatchTopic, error) {
-	wt := mv1.NewWatchTopic(name)
-	ri, err := client.GetResource(wt.GetSelfLink())
+	wt := mv1.NewWatchTopic("")
+	ri, err := client.GetResource(fmt.Sprintf("%s/%s", wt.GetKindLink(), name))
```
jcollins-axway, line 69: This seems to be a 1 for 1 change, is there a benefit of one way vs the other?

sbolosan (dismissed):

sbolosan (approved):

jcollins-axway (approved):

PR #424, opened 2022-05-24
APIGOV-22760 - bubble up error but do not kill the agent

sbolosan:
1.  bubble up the agent to .Migrate so the real error is exposed
2.  do not kill the agent, try to do the other apiservices if they exist

@jcollins-axway @tsjohns9 - hopefully i didn't miss the mark on this one.  Dave has an api that didn't have a spec value.  So the migration caught that and bubbled it up.  But it was getting eaten by the last error and that was being displayed.  Beyond that, if the first of many apiservices failed, we exited.  Removed return from the loop to try all the apiservices before returning and do not kill the agent as part of the migration.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ replace (
 )
 
 retract ( // errored versions
+	[v1.1.21, v1.1.23]
 	v1.1.16
 	[v1.1.4, v1.1.9]
-	[v1.1.21, v1.1.23] 
 )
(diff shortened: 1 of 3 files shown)


jcollins-axway:
That will log the error, but do we want to stop the agent in cases like this? Should we just skip these revisions?

sbolosan:
@jcollins-axway  i don't think i stop it... i removed the return from the cach func

jcollins-axway:
My bad, yes, see that now

jcollins-axway (approved):

vivekschauhan (approved):

PR #426, opened 2022-05-25
APIGOV-22757 - set lower limit polling intervals to 30s

sbolosan:
1. Validate polling interval time durations
2. If duration is over 30, NOP
3. If duration is under 30, warn that the config is under 30s and we are setting the config to 30s

### DIFF
--- a/pkg/cmd/properties/properties.go
+++ b/pkg/cmd/properties/properties.go
@@ -288,20 +288,30 @@ func (p *properties) StringFlagValue(name string) (bool, string) {
 }
 
 func (p *properties) DurationPropertyValue(name string) time.Duration {
+	lowerLimit, _ := time.ParseDuration("30s")
+
 	s := p.parseStringValue(name)
 	d, _ := time.ParseDuration(s)
 
-	// Get seconds duration value and check if duration is less than 30s
-	if d.Seconds() < 30 {
-		log.Warnf("config %s has been reset from %s to the lower limit duration of 30s. Please update this value accordingly in your configs", name, s)
-		// Set duration to the lower limit of 30s
-		d, _ = time.ParseDuration("30s")
+	// Get config value and check if duration is less than 30s
+	if d < time.Second*30 {
+		flagName := p.nameToFlagName(name)
+		flag := p.rootCmd.Flag(flagName)
+
+		// since config value is < 30s, get duration default value
+		defaultValue := flag.DefValue
+		d, _ = time.ParseDuration(defaultValue)
+
+		if d > time.Second*30 {
+			// if defaultValue is > 30s, then just set the lower limit
+			d = lowerLimit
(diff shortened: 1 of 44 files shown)


jcollins-axway (dismissed):

pkg/cmd/properties/properties.go
```diff
@@ -291,6 +291,13 @@ func (p *properties) DurationPropertyValue(name string) time.Duration {
 	s := p.parseStringValue(name)
 	d, _ := time.ParseDuration(s)
 
+	// Get seconds duration value and check if duration is less than 30s
+	if d.Seconds() < 30 {
+		log.Warnf("config %s has been reset from %s to the lower limit duration of 30s. Please update this value accordingly in your configs", name, s)
+		// Set duration to the lower limit of 30s
+		d, _ = time.ParseDuration("30s")
```
vivekschauhan, line 298: Instead of setting to lower limit may be we could use the default values.

sbolosan: We would only set if it's below 30 seconds right? So if default is 60s, we leave it alone. I didn't change any defaults. Should I look at any defaults and change to 30s? This only looks at what the users set in their config. Or maybe I'm missing something.

tsjohns9 (approved):

pkg/cmd/properties/properties.go
```diff
@@ -288,13 +288,30 @@ func (p *properties) StringFlagValue(name string) (bool, string) {
...
+	// Get config value and check if duration is less than 30s
+	if d < time.Second*30 {
+		flagName := p.nameToFlagName(name)
+		flag := p.rootCmd.Flag(flagName)
+
+		// since config value is < 30s, get duration default value
+		defaultValue := flag.DefValue
+		d, _ = time.ParseDuration(defaultValue)
+
+		if d > time.Second*30 {
```
jcollins-axway, line 305: Can you set lowerLimit := time.Second * 30 and additionally use that var in the 2 conditionals?

sbolosan: Yes 🤦‍♂️

pkg/cmd/properties/properties.go
```diff
@@ -288,13 +288,30 @@ func (p *properties) StringFlagValue(name string) (bool, string) {
...
 	d, _ := time.ParseDuration(s)
 
+	// Get config value and check if duration is less than 30s
+	if d < time.Second*30 {
+		flagName := p.nameToFlagName(name)
+		flag := p.rootCmd.Flag(flagName)
+
+		// since config value is < 30s, get duration default value
+		defaultValue := flag.DefValue
+		d, _ = time.ParseDuration(defaultValue)
```
jcollins-axway, line 303: this can be a single line `d, _ = time.ParseDuration(flag.DefValue)`

jcollins-axway (approved):

PR #428, opened 2022-05-25
APIGOV-22682 - update for access request provisioning data

jcollins-axway:


sbolosan (approved):

vivekschauhan (approved):

pkg/agent/discoverycache.go
```diff
@@ -126,7 +126,7 @@ func (j *discoveryCache) updateAPIServiceInstancesCache() {
 	defer j.instanceCacheLock.Unlock()
 
 	query := map[string]string{
-		apic.FieldsKey: apiServerFields,
+		apic.FieldsKey: apiServerFields + "," + defs.Spec,
```
tsjohns9, line 129: I removed queries in the discovery cache

pkg/agent/handler/accessrequest.go
```diff
@@ -20,22 +20,24 @@ const (
...
 // NewAccessRequestHandler creates a Handler for Access Requests
 func NewAccessRequestHandler(prov arProvisioner, cache agentcache.Manager, client client) Handler {
 	return &accessRequestHandler{
-		prov:   prov,
-		cache:  cache,
-		client: client,
+		prov:          prov,
+		cache:         cache,
+		client:        client,
+		encryptSchema: encryptSchema,
```
tsjohns9, line 40: Where does encryptSchema come from?

pkg/agent/handler/accessrequest.go
```diff
@@ -56,6 +58,11 @@ func (h *accessRequestHandler) Handle(ctx context.Context, meta *proto.EventMeta
 		return nil
 	}
 
+	// add or update the cache with the access request
+	if action == proto.Event_CREATED || action == proto.Event_UPDATED {
```
tsjohns9, line 62: How come the resource gets cached before processing?

pkg/agent/handler/accessrequest.go
jcollins-axway: Its the same that the credential uses

pkg/agent/handler/accessrequest.go
jcollins-axway: Want to make sure our cache has the latests, on an update event we may not have to process anything but want the cache updated

tsjohns9 (approved):

vivekschauhan (approved):

PR #430, opened 2022-05-27
APIGOV-22791 - changes from QA testing

sbolosan:
Kickback from testing

1. handle usage reporting when set/unset for DA and TA respectively
2. change logic to see if default actually equals lower limit (of 30s)


sbolosan:
The check for nil re : usage reporting is quick and dirty.  If you find all the references GetUsageReportingConfig, you'll see that this call is being made in DA paths.  Maybe we need to rethink these paths... Right now, I just pass back and empty UsageReportingConfig

jcollins-axway (approved):

PR #440, opened 2022-06-02
APIGOV-21771 - fix race condition in creating subresources

dfeldick:


sbolosan:
so, you actually are seeing race conditions?

dfeldick:
> so, you actually are seeing race conditions?

Yes. If you edit the makefile locally and add --race to the go test lines, you will see race conditions all over. Well, it's a lot better now, but the jobs stuff still is a mess. That's next.

dfeldick:
So, in createSubResource in apic/client.go, there is a loop on line 701 going thru each sub. Each loop starts a go func. Of course, that func will run in the background when it wants and however long it takes. Within that func it is setting a value for "instanceBytes" and "err". Now it loops again with the next sub. Creates another func, which again sets instanceBytes. But since you have no idea which func is going to be writing the data when, the 2nd loop could in theory I guess set the value before the first loop did. Now, maybe for this it doesn't really matter because nobody is actually reading instanceBytes until after the loop. But, the point is that multiple people in multiple threads are writing the same var without a lock... at least that's my understanding from the error I see.

jcollins-axway (approved):

sbolosan:
hmmm... if i have cycles tomorrow, i will look at it.  I did this migration stuff and I thought I did well to protect it.  My assumptions is that the go routine stays within its "own"...  so i didn't think it would be fighting for the same access to a func.

sbolosan:
i know the lock will protect it, but its just for my understanding

PR #442, opened 2022-06-06
APIGOV-22823 - add consumer info to traceability event

sbolosan:


pkg/transaction/eventgenerator.go
```diff
@@ -191,6 +195,57 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
...
+
+	// get subscription info
+	subscription := transutil.GetSubscription(cacheManager, accessRequest)
+	subscriptionID := transutil.GetSubscriptionID(subscription)
+	subscriptionName := subscription.Name
+
+	application := &Application{
+		ID:   unknown,
+		Name: unknown,
+	}
```
jcollins-axway, line 222: Should the subscription data also be init'd to unknown?

sbolosan:
@jcollins-axway @vivekschauhan  set to draft
1. refactored util.  have changes in the agents for these
2. have set to bybass on test because i'm struggling with unit test for event generator
3. moved common code from metrics and eventgenerator to util

aside from the unit test check, can you guys just please look at the core code and make sure you don't think i may have missed something.

thanks!

pkg/transaction/eventgenerator.go
```diff
@@ -191,6 +195,57 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
...
+	cacheManager := agent.GetCacheManager()
+	appName := ""
+
+	if summaryEvent.TransactionSummary.Application != nil {
+		appName = summaryEvent.TransactionSummary.Application.Name
+	}
+
+	// get proxy information
+	apiID := summaryEvent.TransactionSummary.Proxy.ID
+	stage := summaryEvent.TransactionSummary.Proxy.Stage
```
jcollins-axway, line 209: Should we check to make sure all these objects are set? Possible NPE?

sbolosan: i've seen the stage come back "" (blank)...  i can check to make sure.. .

pkg/transaction/eventgenerator.go
jcollins-axway: I guess it is, since the func will return unknown by default

pkg/transaction/eventgenerator.go
vivekschauhan: stage can be blank. I guess we can add a nil check for summaryEvent.TransactionSummary and summaryEvent.TransactionSummary.Proxy

pkg/transaction/eventgenerator.go
```diff
@@ -135,6 +136,9 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
 		log.Debug("Found api path in traceability api exceptions list.  Ignore transaction event")
 		return events, nil
 	}
+	if util.IsNotTest() {
+		summaryEvent.TransactionSummary.ConsumerDetails = e.getConsumerInfo(summaryEvent)
```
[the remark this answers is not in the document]
vivekschauhan: Agree, for test we can setup the cache with consumer specific resources.

tsjohns9 (dismissed):

jcollins-axway:
> @vivekschauhan @jcollins-axway depending on certain objects being nil, i just return an empty ConsumerDetails... or maybe i should not return anything at all?

Maybe return nil that way the consumer object on the event will not be added at all when we serialize to JSON

sbolosan:
> > @vivekschauhan @jcollins-axway depending on certain objects being nil, i just return an empty ConsumerDetails... or maybe i should not return anything at all?
> 
> Maybe return nil that way the consumer object on the event will not be added at all when we serialize to JSON

Good point.. will do!

pkg/transaction/eventgenerator.go
```diff
@@ -135,6 +136,14 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
 		log.Debug("Found api path in traceability api exceptions list.  Ignore transaction event")
 		return events, nil
 	}
+	if util.IsNotTest() {
+		if summaryEvent.TransactionSummary != nil {
+			consumerDetails := e.getConsumerDetails(summaryEvent)
+			if consumerDetails != nil {
+				summaryEvent.TransactionSummary.ConsumerDetails = consumerDetails
```
jcollins-axway, line 143: I think you can still set 

summaryEvent.TransactionSummary.ConsumerDetails = e.getConsumerDetails(summaryEvent)

If nil is returned then it will not be added to the event sent to condor with omitempty being here

ConsumerDetails *ConsumerDetails `json:"consumerDetails,omitempty"`

pkg/transaction/eventgenerator.go
```diff
@@ -191,6 +200,123 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
...
+	if summaryEvent.TransactionSummary.Proxy != nil {
+		apiID = summaryEvent.TransactionSummary.Proxy.ID
+		stage = summaryEvent.TransactionSummary.Proxy.Stage
+		e.logger.
+			WithField("apiID", apiID).
+			WithField("stage", stage).
+			Debug("transaction summary proxy information")
+	} else {
+		return nil
+	}
```
jcollins-axway, line 228: Can you switch the parts of this conditional?

if sumEv....Proxy == nil {
return nil
}

pkg/transaction/eventgenerator.go
```diff
@@ -191,6 +200,123 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
...
+	if managedApp == nil {
+		e.logger.
+			WithField("appName", appName).
+			Debug("could not get managed application by name")
+		return nil
+	}
+	e.logger.
+		WithField("appName", appName).
+		WithField("managed app name", managedApp.Name).
+		Debug("managed application info")
```
jcollins-axway, line 241: Thoughts on making the logger a local var and then continually adding the fields, so subsequent logs has all previous logs? Maybe overkill?

Also, should some of these be trace?

pkg/transaction/eventgenerator.go
```diff
@@ -191,6 +200,123 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
...
+				return nil
+			}
+		}
+		e.logger.
+			WithField("consumer org ID", consumerOrgID).
+			Debug("consumer org ID ")
+	}
+
+	// Update consumer details with Org, Application and Subscription
+	consumerDetails := &ConsumerDetails{
```
jcollins-axway, line 311: just return the object here rather than creating a var?

pkg/transaction/eventgenerator.go
sbolosan: Will do

pkg/transaction/eventgenerator.go
sbolosan: I'll look at it... and yes

pkg/transaction/eventgenerator.go
sbolosan: 👍🏾

pkg/transaction/eventgenerator.go
sbolosan: I left the logging the way it is... maybe overkill?  I can be swayed if the other 2 have thoughts on it.

sbolosan: ...but i did update trace v debug

jcollins-axway (approved):

vivekschauhan (approved):

PR #447, opened 2022-06-10
APIGOV-22657 Sample subscription

tsjohns9:


pkg/traceability/sampling/sample.go
```diff
@@ -20,9 +20,28 @@ func (s *sample) ShouldSampleTransaction(details TransactionDetails) bool {
...
-	if s.config.PerAPI && details.APIID != "" {
-		return s.shouldSampleWithCounter(details.APIID)
+
+	if s.config.PerAPI && s.config.PerSub {
+		if details.SubID != "" {
+			return s.shouldSampleWithCounter(details.SubID)
+		}
+		if details.APIID != "" {
+			return s.shouldSampleWithCounter(details.APIID)
+		}
```
jcollins-axway, line 30: I think we should also have a concatenation of the subid and apiid

pkg/traceability/sampling/sample.go
```diff
@@ -20,9 +21,28 @@ func (s *sample) ShouldSampleTransaction(details TransactionDetails) bool {
 	if hasFailedStatus && s.config.ReportAllErrors {
 		return true
 	}
-	if s.config.PerAPI && details.APIID != "" {
-		return s.shouldSampleWithCounter(details.APIID)
+
+	if s.config.PerAPI && s.config.PerSub {
+		if details.SubID != "" {
+			return s.shouldSampleWithCounter(fmt.Sprintf("%s-%s", details.APIID, details.SubID))
```
tsjohns9, line 27: @jcollins-axway  like this?

jcollins-axway: that looks better, think you can drop the conditional on 29 and just return then

jcollins-axway: that looks better, think you can drop the conditional on 29 and just return then

tsjohns9: We still need to check if there is an api id though?

jcollins-axway: true true, that works

pkg/traceability/sampling/globalsampling.go
```diff
@@ -14,6 +14,7 @@ var agentSamples *sample
 type Sampling struct {
 	Percentage      int  `config:"percentage"    validate:"min=0, max=100"`
 	PerAPI          bool `config:"per_api"`
+	PerSub          bool `config:"per_api"`
```
jcollins-axway, line 17: config setting needs updated

tsjohns9: What about it needs to be updated?

pkg/traceability/sampling/sampling_test.go
```diff
@@ -192,6 +193,64 @@ func TestShouldSample(t *testing.T) {
...
+				"id3": 50, // expect 50
+				"id4": 50, // expect 50
+			},
+			subIDs: map[string]string{
+				"id1": "sub1",
+				"id2": "sub2",
+				"id3": "sub3",
+				"id4": "sub4",
+			},
+			expectedSampled: 200,
```
tsjohns9, line 210: Why is the expectedSampled at 200 when the percent is 50 and the total requests is 200? I copied this from another test which is set up the same way.

pkg/traceability/sampling/globalsampling.go
jcollins-axway: PerSub bool `config:"per_api"`. -> PerSub bool `config:"per_subscription"`

pkg/traceability/sampling/sampling_test.go
jcollins-axway: 4 ids and we are sampling by id. So each of them will be guaranteed 50 of every 100 transactions

jcollins-axway: The way our sampling works, with the counters, the first 50 of each id will be sampled, the next 50 of each id will not be.  So if we then set all ids to send 75 transactions (300 total) we would still expect 200 sampled.

pkg/traceability/sampling/sample.go
```diff
@@ -20,9 +21,28 @@ func (s *sample) ShouldSampleTransaction(details TransactionDetails) bool {
...
+		}
+		if details.APIID != "" {
+			return s.shouldSampleWithCounter(details.APIID)
+		}
+	}
```
jcollins-axway, line 32: I was thinking, in this case where both settings are on.

We probably want to count every apiid, even if subid and apiid are both set.  That way when subid is not set we still have the counts of all transactions for that apiid.

tsjohns9: Ok, what should change here?

jcollins-axway: Something like this, maybe?

```
if s.config.PerAPI && s.config.PerSub {
	if details.SubID != "" {
		s.shouldSampleWithCounter(details.APIID). // keep count of sampling by just api id as well
		return s.shouldSampleWithCounter(fmt.Sprintf("%s-%s", details.APIID, details.SubID))
	}
	if details.APIID != "" {
		return s.shouldSampleWithCounter(details.APIID)
	}
}
```

jcollins-axway (approved):

PR #448, opened 2022-06-13
APIGOV-22861 - refactor migration code

sbolosan:
1. move spec/parser to a more generalized location for easy access (REJECTED)
2. add context to get service name as migration process runs through creation of ARDs/CRDs and apiservice/apiserviceinstance resources
3. update x-agent-details so migration doesn't run after subsequent startups if it was previously ran

Regarding #1.  I went through various iterations to try and move spec/parser out of apic into a common util package.  I even tried to move spec/parser within apic so that servicebody/servicebuilder and migration could get at it.  However, because of definitions being shared across multiple packages, it proved difficult.  Moreover, once I did get something I was almost satisfied with, I found that agents were also using definitions and functions re spec/parser.  I did not bite that bullet.  I don't think biting that bullet of having to change the sdk along with all the agents that called apic package to get definitions and spec/parser funcs was worth it.  So I left it as is...  If majority says bite that bullet, I will :)

### DIFF
--- a/pkg/migrate/marketplacemigration.go
+++ b/pkg/migrate/marketplacemigration.go
@@ -81,7 +81,7 @@ func (m *MarketplaceMigration) Migrate(ri *v1.ResourceInstance) (*v1.ResourceIns
 			// migration ran already
 			m.logger.
 				WithField(string(serviceName), apiSvc).
-				Debugf("marketplace provision migration already completed for service-name")
+				Debugf("marketplace provision migration already completed")
 			return ri, nil
 		}
 	}
--- a/pkg/migrate/marketplacemigration.go
+++ b/pkg/migrate/marketplacemigration.go
@@ -50,6 +50,7 @@ type MarketplaceMigration struct {
 	cache  ardCache
 }
 
+// Migrate -
 func (m *MarketplaceMigration) Migrate(ri *v1.ResourceInstance) (*v1.ResourceInstance, error) {
 	if ri.Kind != mv1a.APIServiceGVK().Kind {
 		return ri, nil
--- a/pkg/migrate/marketplacemigration.go
+++ b/pkg/migrate/marketplacemigration.go
@@ -86,7 +86,7 @@ func (m *MarketplaceMigration) Migrate(ri *v1.ResourceInstance) (*v1.ResourceIns
 		WithField(serviceName, ri.Name).
 		Tracef("perform marketplace provision")
 
-	err = m.updateService(ri, ctx)
+	err = m.updateService(ctx, ri)
 	if err != nil {
 		return ri, fmt.Errorf("migration marketplace provisioning failed: %s", err)
 	}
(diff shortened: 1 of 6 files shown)


jcollins-axway (dismissed):

pkg/migrate/marketplacemigration.go
```diff
@@ -79,7 +79,7 @@ func (m *MarketplaceMigration) Migrate(ri *v1.ResourceInstance) (*v1.ResourceIns
 		completed := details[definitions.MarketplaceMigration]
 		if completed == definitions.MigrationCompleted {
 			// migration ran already
-			m.logger.Debugf("marketplace provision migration ran previously for service-name", apiSvc.Name)
+			m.logger.Debugf("marketplace provision migration ran previously for service-name %s", apiSvc.Name)
```
jcollins-axway, line 82: WithField rather than in log line>

pkg/migrate/marketplacemigration.go
```diff
@@ -79,7 +79,9 @@ func (m *MarketplaceMigration) Migrate(ri *v1.ResourceInstance) (*v1.ResourceIns
 		completed := details[definitions.MarketplaceMigration]
 		if completed == definitions.MigrationCompleted {
 			// migration ran already
-			m.logger.Debugf("marketplace provision migration ran previously for service-name %s", apiSvc.Name)
+			m.logger.
+				WithField(string(serviceName), apiSvc).
+				Debugf("marketplace provision migration ran previously for service-name")
```
jcollins-axway, line 84: marketplace provision migration ran previously for service-name - > marketplace provision migration already completed

?

pkg/migrate/marketplacemigration.go
```diff
@@ -54,7 +65,32 @@ func (m *MarketplaceMigration) Migrate(ri *v1.ResourceInstance) (*v1.ResourceIns
...
+
+	// get x-agent-details and determine if we need to process this apiservice for marketplace provisioning
+	details := util.GetAgentDetails(apiSvc)
+	if len(details) > 0 {
+		completed := details[definitions.MarketplaceMigration]
+		if completed == definitions.MigrationCompleted {
+			// migration ran already
+			m.logger.
+				WithField(string(serviceName), apiSvc).
+				Debugf("marketplace provision migration already completed")
```
tsjohns9, line 84: Can we please remove the log so we don't see this each time we start the agent, and just log when we do migrate a service

pkg/migrate/marketplacemigration.go
```diff
@@ -54,7 +65,32 @@ func (m *MarketplaceMigration) Migrate(ri *v1.ResourceInstance) (*v1.ResourceIns
...
+			return ri, nil
+		}
+	}
+
+	m.logger.
+		WithField(string(serviceName), ri.Name).
```
tsjohns9, line 90: I think serviceName is already a string so you might not need the type conversion

tsjohns9 (approved):

vivekschauhan (approved):

tsjohns9 (approved):

PR #449, opened 2022-06-15
APIGOV-22882 - resource instance prior to converting to Access Request

jcollins-axway:


pkg/agent/cache/accessrequest.go
```diff
@@ -44,11 +45,13 @@ func (c *cacheManager) GetAccessRequestByAppAndAPI(appName, remoteAPIID, remoteA
...
-	accessRequest, _ := c.accessRequestMap.GetBySecondaryKey(appName + ":" + remoteAPIID + ":" + remoteAPIStage)
-	if accessRequest != nil {
-		ri, ok := accessRequest.(*mv1.AccessRequest)
-		if ok {
-			return ri
+	cachedAccessRequest, _ := c.accessRequestMap.GetBySecondaryKey(appName + ":" + remoteAPIID + ":" + remoteAPIStage)
+	if cachedAccessRequest != nil {
+		if ri, ok := cachedAccessRequest.(*v1.ResourceInstance); ok {
+			accessRequest := &mv1.AccessRequest{}
+			if accessRequest.FromInstance(ri) == nil {
```
sbolosan, line 52: If it's nil you wanna return? Or != nil? Slightly confused. No code in front of me.

 Still 🏃🏃🏃

jcollins-axway: FromInstance returns an error only. The object that calls it is loaded to from the resource instance that was sent in.

sbolosan: 👍🏾

sbolosan (approved):
Good call on changing to using resource

vivekschauhan (approved):

PR #455, opened 2022-06-17
[AMPGW-326] Fix discovery cache load issue so that APISeviceInstance cache is loaded before marketplace AccessRequest cache as it requires the APIServiceInstances.

mcollins-axway:


sbolosan (approved):
The quick fix looks good to me. Will wait til either Jason or Trevor get a chance to review 👍🏾

tsjohns9 (approved):

PR #456, opened 2022-06-17
APIGOV-22869 Add job to check the Central health status

tsjohns9:


sbolosan (approved):
I think this somewhat covers the ticket I opened this morning as well (in agent channel)

jcollins-axway (approved):

jcollins-axway:
> I think this somewhat covers the ticket I opened this morning as well (in agent channel)

I think for that ticket we want to be more verbose, when failing on startup, as to the status of each healthcheck.

Something like

Amplify Central..success
API Manager......success
API Gateway......failed

pkg/util/healthcheck/healthchecker.go
```diff
@@ -35,14 +36,14 @@ func init() {
...
-//StartPeriodicHealthCheck - starts a job that runs the periodic healthchecks
+// StartPeriodicHealthCheck - starts a job that runs the periodic health checks
 func StartPeriodicHealthCheck() {
 	interval := defaultCheckInterval
 	if GetStatusConfig() != nil {
 		interval = GetStatusConfig().GetHealthCheckInterval()
 	}
 	periodicHealthCheckJob := &periodicHealthCheck{interval: interval}
-	go periodicHealthCheckJob.Execute()
+	jobs.RegisterIntervalJobWithName(periodicHealthCheckJob, periodicHealthCheckJob.interval, "Periodic Health Check")
```
jcollins-axway, line 46: Since all this does is see if the health check is still good, can't it run every second, for instance? To quickly bring down the jobs when a failure occurs?

jcollins-axway (dismissed):

pkg/util/healthcheck/healthchecker.go
tsjohns9: Is there any reason to have the health check interval config value then?

jcollins-axway: My bad, I was thinking that was your new job.  This is fine.

jcollins-axway (approved):

PR #463, opened 2022-06-25
APIGOV-22993 - Refactor marketplace provisioning migration

sbolosan:
QA tested and found that there were inconsistencies when ARD's and CRD's were added to apiserviceinstances
1. move update service to migrate util
2. remove context (no longer needed since moved to util)
3. update logging
4.  remove scopes for oauth processing
5. honor migration complete flag.  Only call migration on startup.  Only call migration when flag is not "complete"


pkg/agent/provisioning.go
```diff
@@ -100,8 +78,13 @@ func migrateMarketPlace(marketplaceMigration migrate.Migrator, ri *v1.ResourceIn
...
 		var err error
-		log.Tracef("update apiserviceinstances with request definition %s: %s", ri.Kind, ri.Name)
-		_, err = marketplaceMigration.Migrate(svc)
+
+		logger.Tracef("update apiserviceinstances with request definition %s: %s", ri.Kind, ri.Name)
+
+		mig := marketplaceMigration.(*migrate.MarketplaceMigration)
+		migrate.UpdateService(svc, mig)
+
+		// _, err = marketplaceMigration.Migrate(svc)
```
tsjohns9, line 81: dead code?

pkg/migrate/marketplacemigration.go
```diff
@@ -79,382 +62,22 @@ func (m *MarketplaceMigration) Migrate(ri *v1.ResourceInstance) (*v1.ResourceIns
...
 
-	m.logger.
-		WithField(string(serviceName), ri.Name).
+	m.Logger.
+		WithField(serviceName, ri.Name).
 		Tracef("perform marketplace provision")
 
-	err = m.updateService(ctx, ri)
+	UpdateService(ri, m)
+	// err = m.updateService(ctx, ri)
```
tsjohns9, line 77: here too

pkg/agent/provisioning.go
sbolosan: crap.. yes

pkg/migrate/util.go
```diff
@@ -0,0 +1,342 @@
...
+			knownCRDs = append(knownCRDs, policy)
+		}
+	}
+
+	return knownCRDs
+}
+
+func registerAccessRequestDefinition(scopes map[string]string) (*mv1a.AccessRequestDefinition, error) {
+	oauthScopes := make([]string, 0)
+	for scope := range scopes {
```
tsjohns9, line 164: I don't think the scope stuff here is used anymore

sbolosan: right... removing...

pkg/migrate/util.go
```diff
@@ -0,0 +1,342 @@
...
+const (
+	serviceName  = "service-name"
+	instanceName = "instance-name"
+	revisionName = "revision-name"
+)
+
+var apiserviceName string
+
+// UpdateService - gets a list of instances for the service and updates their request definitions.
+func UpdateService(ri *v1.ResourceInstance, m *MarketplaceMigration) error {
```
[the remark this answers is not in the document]
tsjohns9: I see you exported this func so it can be used in provisioning.go. I think you can achieve the same thing by leaving all these funcs on `MarketplaceMigration` and doing your type conversion on the interface to get the `MarketplaceMigration` type and then calling `mig.UpdateService`

tsjohns9:
Looks good

PR #464, opened 2022-06-27
APIGOV-23039 - standardize WithField

sbolosan:
WithField standard, lower case with "-" (dashes)

### DIFF
--- a/pkg/jobs/pool.go
+++ b/pkg/jobs/pool.go
@@ -318,7 +318,7 @@ func (p *Pool) startAll() bool {
 	p.logger.Debug("Checking for all cron jobs to be ready")
 	for _, job := range p.getCronJobs() {
 		if !job.Ready() {
-			p.logger.WithField("job-iD", job.GetID()).Debugf("job is not ready")
+			p.logger.WithField("job-id", job.GetID()).Debugf("job is not ready")
 			return false
 		}
 	}
--- a/pkg/jobs/pool.go
+++ b/pkg/jobs/pool.go
@@ -73,8 +73,8 @@ func (p *Pool) recordJob(job JobExecution) string {
 	p.jobsMapLock.Lock()
 	defer p.jobsMapLock.Unlock()
 	p.logger.
-		WithField("jobID", job.GetID()).
-		WithField("jobName", job.GetName()).
+		WithField("job-id", job.GetID()).
+		WithField("job-name", job.GetName()).
 		Trace("registered job")
 	p.jobs[job.GetID()] = job
 	return job.GetID()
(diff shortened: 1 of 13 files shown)


pkg/jobs/pool.go
```diff
@@ -318,7 +318,7 @@ func (p *Pool) startAll() bool {
 	p.logger.Debug("Checking for all cron jobs to be ready")
 	for _, job := range p.getCronJobs() {
 		if !job.Ready() {
-			p.logger.WithField("jobID", job.GetID()).Debugf("job is not ready")
+			p.logger.WithField("job-iD", job.GetID()).Debugf("job is not ready")
```
dfeldick, line 321: typo. Up to you, but maybe since these strings seem to be used all over they should have had a const definition somewhere?

pkg/transaction/metric/metricscollector.go
```diff
@@ -25,6 +25,12 @@ import (
 	"github.com/Axway/agent-sdk/pkg/util/log"
 )
 
+const (
+	startTimestamp = "start-timestamp"
+	endTimestamp   = "end-timestamp"
+	eventType      = "event-type"
```
dfeldick, line 31: hey, consts here!

pkg/transaction/metric/metricscollector.go
```diff
@@ -167,15 +173,15 @@ func (c *collector) Execute() error {
...
+		WithField(eventType, "usage").
 		Debug("generating usage event")
 
 	c.logger.
-		WithField("startTimestamp", util.ConvertTimeToMillis(c.metricStartTime)).
-		WithField("endTimestamp", util.ConvertTimeToMillis(c.metricEndTime)).
-		WithField("eventType", "metric").
+		WithField(startTimestamp, util.ConvertTimeToMillis(c.metricStartTime)).
+		WithField(endTimestamp, util.ConvertTimeToMillis(c.metricEndTime)).
+		WithField(eventType, "metric").
```
dfeldick, line 184: and here we are using some consts and some inline strings

pkg/transaction/metric/metricscollector.go
sbolosan: i did this because it made sense in this file

pkg/transaction/eventgenerator.go
```diff
@@ -286,8 +286,8 @@ func (e *Generator) getConsumerDetails(summaryEvent LogEvent) *ConsumerDetails {
 		application.ID = appID
 		application.Name = appName
 		e.logger.
-			WithField("application ID", application.ID).
-			WithField("application name", application.Name).
+			WithField("application-id", application.ID).
+			WithField("application-name", application.Name).
```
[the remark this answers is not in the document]
sbolosan: id represents an actual alphanumeric string.  name is the name of the app.

pkg/transaction/metric/metricscollector.go
sbolosan: good point... fixing these too

sbolosan:
If in the future, as we start adding WithField logging, it dictates constants, then we can add that for sure.  Right now, only metricscollector seemed necessary.

dfeldick (approved):

PR #492, opened 2022-07-18
APIGOV-23178 - add product version id

sbolosan:
Spoke with Levan and Romain - add product version id and keep product version name... 
landed on versionName and versionId for the key values.

### DIFF
--- a/pkg/transaction/eventgenerator.go
+++ b/pkg/transaction/eventgenerator.go
@@ -396,9 +396,10 @@ func updateWithProviderDetails(accessRequest *v1alpha1.AccessRequest, managedApp
 	}
 
 	summaryEvent.Product = &models.Product{
-		ID:      unknown,
-		Name:    unknown,
-		Version: unknown,
+		ID:          unknown,
+		Name:        unknown,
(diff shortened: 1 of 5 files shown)


pkg/transaction/eventgenerator.go
```diff
@@ -396,9 +396,10 @@ func updateWithProviderDetails(accessRequest *v1alpha1.AccessRequest, managedApp
 	}
 
 	summaryEvent.Product = &models.Product{
-		ID:      unknown,
-		Name:    unknown,
-		Version: unknown,
+		ID:          unknown,
+		Name:        unknown,
+		Version:     unknown,
```
vivekschauhan, line 401: Can you name it VersionID to be explicit?

vivekschauhan (dismissed):

pkg/transaction/eventgenerator.go
sbolosan: yes :)

vivekschauhan (approved):

jcollins-axway (approved):

PR #494, opened 2022-07-19
APIGOV-23162 - update schema builder

sbolosan:
- SetPropertyOrder (x-axway-order)
- SetAsTextArea (x-axway-widget)

x-axway-order : string array of property names in the order the implementer wants it in
x-axway-widget : if implementer sets text as area, set widget to "textArea"


pkg/apic/provisioning/schemabuilder.go
```diff
@@ -53,17 +64,40 @@ func (s *schemaBuilder) SetDescription(description string) SchemaBuilder {
...
+	propertyOrderSet = true
+	return s
+}
+
 // AddProperty - adds a new subscription schema property to the schema
 func (s *schemaBuilder) AddProperty(property PropertyBuilder) SchemaBuilder {
 	prop, err := property.Build()
 	if err == nil {
 		s.properties[prop.Name] = *prop
+		// If property order wasn't set, add property as they come in
```
sbolosan, line 80: Originally had an order that was weighted.  But now, if property order not set, append them as the property is added.

pkg/apic/provisioning/schemabuilder_test.go
```diff
@@ -47,4 +48,57 @@ func TestSubscriptionSchemaBuilderSetters(t *testing.T) {
...
+			SetEnumValues([]string{"a", "b", "c"})).
+		AddProperty(NewSchemaPropertyBuilder().
+			SetName("name3").
+			SetDescription("description").
+			SetRequired().
+			IsString().
+			SetEnumValues([]string{"a", "b", "c"})).
+		Build()
+
+	assert.Nil(t, err)
```
jcollins-axway, line 77: Are you doing validations elsewhere that things were set properly?

pkg/apic/provisioning/schemabuilder.go
```diff
@@ -75,6 +109,18 @@ func (s *schemaBuilder) Build() (map[string]interface{}, error) {
...
+	// validate property order
+	for _, value := range s.properties {
+		if len(s.propertyOrder) > 0 {
+			// if property is not in the set property order, warn
+			if !inPropertyOrder(value.Name, s.propertyOrder) {
+				log.Warnf("property %s is not found in the property order", value.Name)
+			}
+		}
+
+	}
```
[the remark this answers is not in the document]
sbolosan: added reverse check

pkg/apic/provisioning/schemabuilder_test.go
sbolosan: not really.. i just did a "test" for the reverse order of properties that were not added but property names in the property order.  But, for either of those scenarios, we don't throw any errors, just warns.  So, there is no assertNotNil for error right? Unless I'm misunderstanding some j-ish.

jcollins-axway: I was wondering if you could check the property order in the returned object, currently _, for the correct value.

sbolosan: ah...gotcha!  can do.  thanks!

jcollins-axway: Is it worth checking? I see no other validations check anything but for err.

sbolosan: agreed

pkg/apic/provisioning/propertybuilder.go
```diff
@@ -36,6 +36,7 @@ type propertyDefinition struct {
 	Minimum            *float64                      `json:"minimum,omitempty"`  // We use a pointer to differentiate the "blank value" from a chosen 0 min value
 	Maximum            *float64                      `json:"maximum,omitempty"`  // We use a pointer to differentiate the "blank value" from a chosen 0 max value
 	IsEncrypted        bool                          `json:"x-axway-encrypted,omitempty"`
+	TextArea           string                        `json:"x-axway-widget,omitempty"`
```
vivekschauhan, line 39: May be this can be named as Widget and the SetAsTextArea sets the widget as "textarea"

pkg/apic/provisioning/schemabuilder_test.go
sbolosan: Created some checks and balances in regards to property order

pkg/apic/provisioning/propertybuilder.go
sbolosan: yeah, will do, since the widget can be of type widget

pkg/apic/provisioning/schemabuilder.go
```diff
@@ -1,11 +1,19 @@
...
 
-import "encoding/json"
+import (
+	"encoding/json"
+
+	"github.com/Axway/agent-sdk/pkg/util/log"
+)
+
+// propertyOrderSet - global flag to check if property order was set
+var propertyOrderSet = false
```
jcollins-axway, line 10: This should not be global, it should be part of the schemabuilder struct

sbolosan: done...

jcollins-axway (approved):

PR #498, opened 2022-07-25
APIGOV-23227 - allow implementer the option of setting a lower and upper limit range for a configuration

sbolosan:
1.  Add new options WithXXXLimit to func AddDurationProperty.  It takes a lower limit and upper limit time.Duration range

2. When implemented, a duration property will contain a lower limit and upper limit function.  It will only be added as a property if the lower limit is less than the upper limit.  And neither of the limits are zero.  When added, new property name and flags will be created for this range

3. DOES not remove logic for - qaEnforceDurationLowerLimit = "QA_ENFORCE_DURATION_LOWER_LIMIT"

4. Validation checks to see if the user entered config is between range

5. If it is, NOP

6. If it is not, then warn user that the set config is not within range and pick up DEFAULT setting

7. Update tests (to check warning).  Function does not return error

### DIFF
--- a/pkg/cmd/cmd_test.go
+++ b/pkg/cmd/cmd_test.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/Axway/agent-sdk/pkg/apic/definitions"
+	"github.com/Axway/agent-sdk/pkg/cmd/properties"
 
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
@@ -59,7 +60,6 @@ func assertDurationCmdFlag(t *testing.T, cmd AgentRootCmd, propertyName, flagNam
 type agentConfig struct {
 	bProp                 bool
 	dProp                 time.Duration
-	rProp                 time.Duration
 	iProp                 int
 	sProp                 string
 	sPropExt              string
@@ -874,7 +874,6 @@ func TestRootCmdAgentConfigValidationNotInDurationRange(t *testing.T) {
 				agentValidationCalled: false,
 				bProp:                 rootCmd.GetProperties().BoolPropertyValue("agent.bool"),
 				dProp:                 rootCmd.GetProperties().DurationPropertyValue("agent.duration"),
-				rProp:                 rootCmd.GetProperties().DurationPropertyValue("agent.duration.range"),
 				iProp:                 rootCmd.GetProperties().IntPropertyValue("agent.int"),
 				sProp:                 rootCmd.GetProperties().StringPropertyValue("agent.string"),
 				ssProp:                rootCmd.GetProperties().StringSlicePropertyValue("agent.stringSlice"),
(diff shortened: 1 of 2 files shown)


pkg/cmd/properties/properties.go
```diff
@@ -36,6 +41,7 @@ type Properties interface {
 	AddStringPersistentFlag(name string, defaultVal string, description string)
 	AddStringFlag(name string, description string)
 	AddDurationProperty(name string, defaultVal time.Duration, description string)
+	AddDurationRangeProperty(name string, defaultVal time.Duration, description string, lowerLimit, upperLimit time.Duration)
```
jcollins-axway, line 44: Rather than a new func here can we update the old one to have an optional set of functional arguments?

AddDurationProperty(name string, defaultVal time.Duration, description string, opts func()...)

jcollins-axway (changes requested):

pkg/cmd/properties/properties.go
sbolosan: updated to use optional funcs... thank you for the teach

vivekschauhan (dismissed):

pkg/cmd/properties/properties.go
```diff
@@ -179,15 +203,45 @@ func (p *properties) AddStringSliceProperty(name string, defaultVal []string, de
...
+
 		p.rootCmd.Flags().Duration(flagName, defaultVal, description)
 		p.bindOrPanic(name, p.rootCmd.Flags().Lookup(flagName))
 		p.rootCmd.Flags().MarkHidden(flagName)
 	}
 }
 
+func (p *properties) addDurationRange(lowerLimitVal, upperLimitVal time.Duration) bool {
+	// if either range value == 0, return false
+	if lowerLimitVal == 0 || upperLimitVal == 0 {
```
jcollins-axway, line 234: So this requires each limit to be set to use either? Shouldn't we allow for floor, ceiling, or both?

jcollins-axway: @vivekschauhan

jcollins-axway (changes requested):

pkg/cmd/properties/properties.go
sbolosan: I like the proposal,  makes sense to me. 👍🏾. Thoughts,  @vivekschauhan, I can get with this.

jcollins-axway: @vivekschauhan agreed with this

sbolosan: okee dokee

jcollins-axway (dismissed):

sbolosan:
latest changes validates limits individually if they are set.

pkg/cmd/properties/properties.go
```diff
@@ -317,24 +363,62 @@ func (p *properties) DurationPropertyValue(name string) time.Duration {
...
 	p.addPropertyToFlatMap(name, s)
 	return d
 }
+
+// validateLowerAndUpperLimits - check limits individually based on if greater than zero
+func (p *properties) validateLowerAndUpperLimits(duration time.Duration, flagName string) bool {
+	lowerLimitFlag := p.rootCmd.Flag(fmt.Sprintf(lowerLimitName, flagName))
+	upperLimitFlag := p.rootCmd.Flag(fmt.Sprintf(upperLimitName, flagName))
+
+	var lowerLimitDuration time.Duration
```
vivekschauhan, line 396:
```suggestion
	if lowerLimitFlag != nil {
		lowerLimitDuration, _ := time.ParseDuration(lowerLimitFlag.Value.String())
		// validate that lower limit is greater than zero and less than configured duration
		if lowerLimitDuration > 0 && lowerLimitDuration > duration {
			log.Warnf(lowerLimitFlag.Usage, duration, lowerLimitDuration, flagName)
			return false
		}
	}
	if upperLimitFlag != nil {
		upperLimitDuration, _ := time.ParseDuration(upperLimitFlag.Value.String())
		// validate that upper limit is greater than zero and greater than configured duration
	        if upperLimitDuration > 0 && upperLimitDuration < duration {
		        log.Warnf(upperLimitFlag.Usage, duration, upperLimitDuration, flagName)
		        return false
	        }
	}
```

pkg/cmd/cmd_test.go
```diff
@@ -858,3 +859,116 @@ func newTestServer() *httptest.Server {
...
+
+	// duration range is invalid.  Upper limit is less than configured value
+	/*
+		{"level":"warning","message":"value 30s is higher than the supported higher limit (20s) for configuration agentDuration","time":"2022-07-26T11:15:48-07:00"}
+		{"level":"warning","message":"config agentDuration has been set to the the default value of 25s.","time":"2022-07-26T11:15:48-07:00"}
+	*/
+	rootCmd.GetProperties().AddDurationProperty("agent.duration", 25*time.Second, "Agent Duration Property", properties.WithUpperLimit(20*time.Second))
+	_ = rootCmd.Execute()
+
+}
```
jcollins-axway, line 974: Seems all of these funcs can be combined into a single test func with multiple test cases.

pkg/cmd/cmd_test.go
```diff
@@ -858,3 +859,81 @@ func newTestServer() *httptest.Server {
...
+	testCases := []struct {
+		name            string
+		defaultDuration time.Duration
+		description     string
+		lowerLimit      time.Duration
+		upperLimit      time.Duration
+	}{
+		{
+			// valid range
+			name:            "agent.duration",
```
jcollins-axway, line 873: can you give the tests better names?

jcollins-axway (approved):

PR #506, opened 2022-08-03
APIGOV-23280 - check central connectivity before calling agent run

sbolosan:


sbolosan:
@tsjohns9  cc @jcollins-axway , can you read https://jira.axway.com/browse/APIGOV-22981 and let me know if it jogs any of your memories... this will work, but i can't find the scenario where it will get to this point, where this check needs to be made.  Any bad URL or central being down is caught before this logic update

pkg/cmd/root.go
```diff
@@ -347,6 +347,14 @@ func (c *agentRootCommand) run(cmd *cobra.Command, args []string) (err error) {
 				properties.SetAliasKeyPrefix(c.agentName)
 				log.SetIsLogP()
 			}
+
+			// Verify we can connect to central.  If we cannot, then kill the agent
+			_, err := agent.GetCentralClient().GetTeam((map[string]string{}))
+			if err != nil {
+				log.Error("Stopping agent - error connecting to Amplify Central. Check docs.axway.com for more info on this error code")
```
jcollins-axway, line 354: nix "code" in this error

tsjohns9: I remember this ticket. The agents used to wait to discover until the central health check passed, and we removed that check from the agents. The main health check now is the "Stream Client" or "Poll Client" connection to central since that is the main connection. I don't know if this function you added is running after all the sdk setup or not. But, I feel like the agents shouldn't start if there was an issue with the chosen connection mode.

tsjohns9: If the stream/poll client goes down, what is the behavior of the agent? Does it try to discover or publish data to central even though the client is down? Or does the agent not do anything until the client is back up?

jcollins-axway: Maybe the root cause for the ticket is not the Central healthcheck but the fact that we no longer call the dataplane healthchecks on startup. For instance iv the Axway Gateway is down when we start up the agent will not fail, although the agent should have already registered its own healthcheck method. Perhaps the SDK should call all healthcheck methods prior to calling the agent run/main func?

tsjohns9 (approved):

PR #511, opened 2022-08-10
INT - Fix for setting credential data for external IDP client

vivekschauhan:


sbolosan (dismissed):

jcollins-axway (approved):

jcollins-axway:
@vivekschauhan can we set data to an empty map if it comes back as nil from encryptSchema?

I missed it.

jcollins-axway (approved):

PR #516, opened 2022-08-17
APIGOV-23402 migrate api service instances to remove finalizers

tsjohns9:
- Remove finalizers on existing instances
- Do not add a finalizer to new instances

### DIFF
--- a/pkg/apic/service.go
+++ b/pkg/apic/service.go
@@ -142,26 +142,9 @@ func (c *ServiceClient) checkReferencesToAccessRequestDefinition(ard string) int
 
 // DeleteAPIServiceInstanceWithFinalizers deletes an api service instance in central, handling finalizers
 func (c *ServiceClient) DeleteAPIServiceInstanceWithFinalizers(ri *v1.ResourceInstance) error {
+	log.DeprecationWarningReplace("DeleteAPIServiceInstanceWithFinalizers", "DeleteAPIServiceInstance")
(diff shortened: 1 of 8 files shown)


pkg/apic/service.go
```diff
@@ -142,27 +142,10 @@ func (c *ServiceClient) checkReferencesToAccessRequestDefinition(ard string) int
 
 // DeleteAPIServiceInstanceWithFinalizers deletes an api service instance in central, handling finalizers
 func (c *ServiceClient) DeleteAPIServiceInstanceWithFinalizers(ri *v1.ResourceInstance) error {
+	log.DeprecationWarningReplace("DeleteAPIServiceInstanceWithFinalizers", "DeleteAPIServiceInstance")
```
tsjohns9, line 145: I'm guessing we still need to keep this method on the ServiceClient, right? If not I can remove it.

jcollins-axway: Thinkint about it I am quite sure the SDK was the only thing to call it. Unless @vivekschauhan thinks different I am going to say blow it away.

jcollins-axway (dismissed):

vivekschauhan (approved):

pkg/apic/service.go
tsjohns9: I will remove it

jcollins-axway (approved):

PR #518, opened 2022-08-17
APIGOV-34567 - enhance error and mitigation

sbolosan:
@jcollins-axway @vivekschauhan - took a stab at this... too verbose?  Wrong mitigation tactic?

### DIFF
--- a/pkg/agent/handler/accessrequest.go
+++ b/pkg/agent/handler/accessrequest.go
@@ -107,7 +107,9 @@ func (h *accessRequestHandler) onPending(ctx context.Context, ar *management.Acc
 
 	// check the application status
 	if app.Status.Level != prov.Success.String() {
-		err = fmt.Errorf("error can't handle access request when application is not yet successful")
+		errStr1 := "your application is currently being processed and has not yet been created on the corresponding Gateway. "
+		errStr2 := "Therefore, your access request cannot be completed. Please refresh.  If the error persists, you can try to delete the application, re-create it and request access again."
+		err = fmt.Errorf("%s%s", errStr1, errStr2)
 		h.onError(ctx, ar, err)
 		return ar
 	}


sbolosan:
https://teams.microsoft.com/l/message/19:3107e9cb6d0a4253b808d79b7345d478@thread.skype/1660726018161?tenantId=300f59df-78e6-436f-9b27-b64973e34f7d&groupId=1501507f-8dd2-495a-a7ac-308a62df8659&parentMessageId=1660672719877&teamName=Amplify%20API%20Platform&channelName=Agents&createdTime=1660726018161&allowXTenantAccess=false

jcollins-axway:
> @jcollins-axway @vivekschauhan - took a stab at this... too verbose? Wrong mitigation tactic?

It's more detailed, but I think we need the UI to have a way to retry an access request. The delete of the app and requesting access again is a bit much.

Maybe say what went wrong only and when we get a retry we van be more verbose about solutions.

Do we default to, contact your administrator??? Or similar.

pkg/agent/handler/accessrequest.go
```diff
@@ -107,7 +107,9 @@ func (h *accessRequestHandler) onPending(ctx context.Context, ar *management.Acc
 
 	// check the application status
 	if app.Status.Level != prov.Success.String() {
```
sbolosan, line 109: @jcollins-axway , well the code only checks for Success... so the other options are Error or Pending.  I think in Leor's case it was "stuck" in pending... i think because he had health check issues at the time he tried to request access.  So, unless we check the status level for error or processing?

jcollins-axway: Can we have multiple checks then? If the app is Error return error saying app creation failed, only mitigation is to recreate the app/access request.

If pending then we have another issue where the app was not handled and not able to update its status, that is an interesting scenario that should not happen, but seems to have in this case.

sbolosan: break it up... yep, can definitely do!

jcollins-axway: may need something similar for credentials?

vivekschauhan (approved):

PR #519, opened 2022-08-18
Ampgw 879

wkolsi:
We have added this flag which is true by default, so that we can on the GovernanceAgent side manage the status, but we keeping the periodic status update.

### DIFF
--- a/pkg/config/agentfeaturesconfig.go
+++ b/pkg/config/agentfeaturesconfig.go
@@ -19,24 +19,24 @@ type AgentFeaturesConfig interface {
 type AgentFeaturesConfiguration struct {
 	AgentFeaturesConfig
 	IConfigValidator
-	ConnectToCentral         bool              `config:"connectToCentral"`
-	ProcessSystemSignals     bool              `config:"processSystemSignals"`
-	VersionChecker           bool              `config:"versionChecker"`
-	PersistCache             bool              `config:"persistCache"`
-	MarketplaceProvisioning  bool              `config:"marketplaceProvisioning"`
-	ExternalIDPConfig        ExternalIDPConfig `config:"idp"`
-	EnableAgentStatusUpdates bool              `config:"enableAgentStatusUpdates"`
+	ConnectToCentral        bool              `config:"connectToCentral"`
+	ProcessSystemSignals    bool              `config:"processSystemSignals"`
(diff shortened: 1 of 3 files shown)


jcollins-axway (dismissed):

pkg/config/agentfeaturesconfig.go
```diff
@@ -24,6 +25,7 @@ type AgentFeaturesConfiguration struct {
 	PersistCache            bool              `config:"persistCache"`
 	MarketplaceProvisioning bool              `config:"marketplaceProvisioning"`
 	ExternalIDPConfig       ExternalIDPConfig `config:"idp"`
+	SdkAgentStatusUpdates   bool              `config:"sdkAgentStatusUpdates"`
```
vivekschauhan, line 28: Can this be changed to enableAgentStatusUpdates?

jcollins-axway (approved):

vivekschauhan (approved):

pkg/config/agentfeaturesconfig.go
wkolsi: Can we rename it as agentStatusUpdates  without the "enable" to be consistent with all the other flags like (versionChecker, marketplaceProvisioning, connectToCentral , ...)?

pkg/config/agentfeaturesconfig.go
```diff
@@ -90,17 +99,19 @@ func AddAgentFeaturesConfigProperties(props properties.Properties) {
 	props.AddBoolProperty(pathVersionChecker, true, "Controls whether the agent SDK version checker will be enabled or not")
 	props.AddBoolProperty(pathPersistCache, false, "Controls whether the agent SDK will persist agent cache or not")
 	props.AddBoolProperty(pathMarketplaceProvisioning, false, "Controls whether the agent should handle Marketplace Subscriptions or not")
+	props.AddBoolProperty(pathEnableAgentStatusUpdates, false, "Controls whether the agent should manage the status update or not")
```
vivekschauhan, line 102: The default here should be true

vivekschauhan (approved):

pkg/config/agentfeaturesconfig.go
jcollins-axway: I agree that we should drop enable from the name.

vivekschauhan (approved):

PR #530, opened 2022-09-01
APIGOV-23542 - fixes for instance counting agains services in cache

jcollins-axway:
- Fix persistent cache to correctly save instance count to disk
- When a new service is added to the cache check to see if any instances linking to that service exist in cache


sbolosan (approved):
team work makes the dream work... thanks @jcollins-axway  @vivekschauhan

vivekschauhan (approved):

PR #541, opened 2022-09-14
APIGOV-23498 - sonar cleanup

dfeldick:
Some of these changes are very simple. Others were refactoring consisting of taking blocks of code within ifs or for loops and refactoring it out into functions to eliminate the indent depth.


jcollins-axway:
everything looks good, but would like some testing to be done before merging to main

sbolosan (approved):

PR #546, opened 2022-09-23
APIGOV-23709 - directly execuate discovered functions

sbolosan:
In testing migration the discovered func (called by go routines) were not being called in the order it needed to be.  Refactored execution by removing go routines and calling functions explicitly

### DIFF
--- a/pkg/agent/discoverycache.go
+++ b/pkg/agent/discoverycache.go
@@ -2,7 +2,6 @@ package agent
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


sbolosan:
APIServiceInstance func was being called before APIService func was done being executed.  This caused the instance to not be there at the time the instance validator was done.. and subsequently, the apiservice was being deleted.

jcollins-axway (approved):

PR #549, opened 2022-09-24
APIGOV-23714 - add logging for QAA

sbolosan:
Possible pitfalls where logging could be used for instance validation deleting apiservice when it's not supposed to

### DIFF
--- a/pkg/agent/cache/apiservice.go
+++ b/pkg/agent/cache/apiservice.go
@@ -152,8 +152,14 @@ func (c *cacheManager) DeleteAPIService(key string) error {
 	return err
 }
 
-func formatKey(svcName string) string {
-	return fmt.Sprintf("count-%v", svcName)
+// FormatKey -
+func (c *cacheManager) FormatKey(svcName string) string {
+	formatKey := fmt.Sprintf("count-%v", svcName)
+	c.logger.
+		WithField("svc-name", svcName).
+		WithField("format-key", formatKey).
+		Debug("format key to be stored for instance count")
+	return formatKey
 }
 
 func (c *cacheManager) addToServiceInstanceCount(apiID, primaryKey string) error {
@@ -162,10 +168,14 @@ func (c *cacheManager) addToServiceInstanceCount(apiID, primaryKey string) error
 		svc = c.GetAPIServiceWithAPIID(apiID)
 		if svc == nil {
 			// can't increment a count for a service we can't find
+			c.logger.
+				WithField("primary-key", primaryKey).
+				WithField("api-id", apiID).
+				Debug("cannot increment a count for a service we cannot find")
 			return nil
 		}
 	}
-	key := formatKey(svc.Name)
+	key := c.FormatKey(svc.Name)
 
 	svcCountI, _ := c.instanceCountMap.Get(key)
 	svcCount := apiServiceToInstanceCount{}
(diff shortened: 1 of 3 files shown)


dfeldick (approved):

sbolosan:
@jcollins-axway @vivekschauhan @tsjohns9
if you get a chance to see where there are opportunities to add logging to triage X's issue.
Right now, when she turns on marketplace from false to true, 1 of her 2 apiservices is getting deleted. I put in a fix previously in discoverycache (today) to resolve an issue where the discovered funcs were getting executed out of order. So i removed the go routines and explicitly just called the funcs

func (dc *discoveryCache) executeDiscoveryFuncs(discoveryFuncs []discoverFunc) error { for _, fun := range discoveryFuncs { err := fun() if err != nil { return err } } return nil }

this seems to work for me. I cannot get it fail. Previous to this fix, I was able to get it to fail.

But for Xylina, running the same code (and configs), she still gets the scenario where one of the two apiservices are getting deleted.
So, looking for areas where we might be out of order, or areas where it thinks the service is no longer there and therefor the instancevalidator is deleting it.

jcollins-axway:
> @jcollins-axway @vivekschauhan @tsjohns9 if you get a chance to see where there are opportunities to add logging to triage X's issue. Right now, when she turns on marketplace from false to true, 1 of her 2 apiservices is getting deleted. I put in a fix previously in discoverycache (today) to resolve an issue where the discovered funcs were getting executed out of order. So i removed the go routines and explicitly just called the funcs
> 
> func (dc *discoveryCache) executeDiscoveryFuncs(discoveryFuncs []discoverFunc) error { for _, fun := range discoveryFuncs { err := fun() if err != nil { return err } } return nil }
> 
> this seems to work for me. I cannot get it fail. Previous to this fix, I was able to get it to fail.
> 
> But for Xylina, running the same code (and configs), she still gets the scenario where one of the two apiservices are getting deleted. So, looking for areas where we might be out of order, or areas where it thinks the service is no longer there and therefor the instancevalidator is deleting it.

Maybe the best approach is to revert the change I made where the cache is keeping the count of Instances to Service. Or, rather than use the count have the methods that are returning the counts go through and do the counting themselves everytime.

```go
func (c *cacheManager) GetAPIServiceInstanceCount(svcName string) int {
	svc := c.GetAPIServiceWithName(svcName)

	// get apiid and primary key
	apiID, _ := util.GetAgentDetailsValue(svc, defs.AttrExternalAPIID)
	primaryKey, _ := util.GetAgentDetailsValue(svc, defs.AttrExternalAPIPrimaryKey)

	count := 0
	for _, k := range c.instanceMap.GetKeys() {
		item, _ := c.instanceMap.Get(k)
		inst, ok := item.(*v1.ResourceInstance)
		if !ok {
			continue
		}
		instAPIID, _ := util.GetAgentDetailsValue(inst, defs.AttrExternalAPIID)
		instPrimary, _ := util.GetAgentDetailsValue(inst, defs.AttrExternalAPIPrimaryKey)
		if apiID == instAPIID || primaryKey == instPrimary {
			c.addToServiceInstanceCount(apiID, primaryKey)
		}
	}

	return count
}
```


Then all the stuff I had added to count and keep the count in the cache can be removed.

sbolosan:
So, we may need to step back and look at the order the funcs are being ran again.  @jcollins-axway your suggestion isn't working fully... now I can see more often then not the instance validator deleting all my apiservices now.  I haven't dug down deeper yet, but this is happening even before i turn marketplace migration on.  So, i'm going to revert. your suggestion and see if I can look at how the discovered funcs are being called.

sbolosan:
I can update and give X a newer build to try this morning.  Thank you

PR #553, opened 2022-09-27
APIGOV-44444 - revert sampling key

sbolosan:
revert from code smell for now.
values are not get set properly when calling into the func for sampling key.  Will address later


jcollins-axway (approved):

ingo:
I was very excited about this ticket number until I realized it wasn't real.

sbolosan:
@ingo I know there's something awesome about all the 4's

PR #555, opened 2022-09-28
APIGOV-SB001 - more revert for sonar fixes

sbolosan:
Unfortunately, more reverts from sonar fixes.  Found that pointers weren't getting passed correctly and this request definitions were not getting set properly.  Will have to circle back and fix the sonar code smells at a later time


sbolosan:
Regression tested again... with the revert, the request definitions are now correctly assigned

jcollins-axway (approved):

PR #556, opened 2022-09-28
APIGOV-23745 - update migration in provisioning part only

sbolosan:
1. remove check to agent details from provisioning.  Leave the check only in the marketplacemigration

2. Only run migration if the resource instance doesn't exist.  If it does exist, we know that migration has already ran

### DIFF
--- a/pkg/agent/provisioning.go
+++ b/pkg/agent/provisioning.go
@@ -54,12 +54,7 @@ func runMarketplaceMigration(ri *v1.ResourceInstance, marketplaceMigration migra
 		existingRI, _ = agent.cacheManager.GetCredentialRequestDefinitionByName(ri.Name)
 	}
 
-	if existingRI == nil {
-		if marketplaceMigration != nil {
-			return true
-		}
-	}
-	return false
+	return existingRI == nil && marketplaceMigration != nil
 }
 
 // migrateMarketPlace -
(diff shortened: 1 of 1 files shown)


jcollins-axway (approved):

pkg/agent/provisioning.go
```diff
@@ -35,13 +36,32 @@ func createOrUpdateDefinition(data v1.Interface, marketplaceMigration migrate.Mi
...
+
+	switch ri.Kind {
+	case management.AccessRequestDefinitionGVK().Kind:
+		existingRI, _ = agent.cacheManager.GetAccessRequestDefinitionByName(ri.Name)
+	case management.CredentialRequestDefinitionGVK().Kind:
+		existingRI, _ = agent.cacheManager.GetCredentialRequestDefinitionByName(ri.Name)
+	}
+
+	if existingRI == nil {
+		if marketplaceMigration != nil {
```
jcollins-axway, line 58: These can be in one conditional

jcollins-axway: in fact you can

return existingRI == nil && marketplaceMigration != nil

jcollins-axway (approved):

jcollins-axway (approved):

tsjohns9 (approved):

PR #564, opened 2022-10-05
APIGOV-23414 update credential state when provisioning fails

tsjohns9:


pkg/agent/handler/credential.go
```diff
@@ -356,6 +356,8 @@ func (h *credentials) provisionPostProcess(status prov.RequestStatus, credential
 			cred.Spec.State.Rotate = false
 			h.client.UpdateResourceInstance(cred)
 		}
+	} else if cred.State.Name == "" {
```
sbolosan, line 359: So setting State.Name to v1.Inactive only happens if status level is not success?

jcollins-axway: Not success and no existing state, otherwise the state of that cred stays as is.

jcollins-axway (approved):

sbolosan (approved):

sbolosan (approved):

pkg/agent/handler/credential.go
tsjohns9: It was getting set for a success. Now it gets set when there is an error too.

PR #567, opened 2022-10-12
APIGOV-23841 - add trace logging for flush out harvester connect issue

sbolosan:
@jcollins-axway @tsjohns9 - I feel like some of the trace logging is lame, like doesn't really do anything except leave bread crumbs to the real error.  And maybe that's what we want.  Let me know if you need me to add more/take away ...  thanks.

### DIFF
--- a/pkg/agent/poller/client.go
+++ b/pkg/agent/poller/client.go
@@ -11,7 +11,6 @@ import (
 	"github.com/Axway/agent-sdk/pkg/harvester"
 	"github.com/Axway/agent-sdk/pkg/util/errors"
 	hc "github.com/Axway/agent-sdk/pkg/util/healthcheck"
-	"github.com/Axway/agent-sdk/pkg/util/log"
 	"github.com/Axway/agent-sdk/pkg/watchmanager/proto"
 )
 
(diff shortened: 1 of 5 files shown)


jcollins-axway (changes requested):
This does not need to be merged into the SDK, until we discover the actual issue. We can give a feature build for just the logging while we investigate.

jcollins-axway:
I am fine with all the logging, verbose or not, as it is trace level and meant only for developers

sbolosan:
@jcollins-axway right.. i will close this after we get what we need.  I already pinged Ethan and that's where I was going... was going to ask if he is ok taking a build from me

pkg/agent/poller/client.go
```diff
@@ -63,16 +64,22 @@ func NewPollClient(
 
 // Start the polling client
 func (p *PollClient) Start() error {
+	logger := log.NewFieldLogger().
+		WithComponent("pollExecutor").
```
tsjohns9, line 68: should be PollClient

tsjohns9 (dismissed):

pkg/agent/poller/client.go
sbolosan: oops, you're right... copy paste

tsjohns9 (dismissed):

sbolosan:
Reverted most of the traces.  Now on startup, using ticker and ticker timeout to check hc status.
ticker - check every 5 seconds
tickerTImeout - exhaust after 5 minutes...

fail if 5 mins exhaust

jcollins-axway:
I think some of the logging that was added may be beneficial. It may be verbose but in some cases nice to see when the code gets in and out of functions.

jcollins-axway (dismissed):

sbolosan:
@jcollins-axway , i have the trace version sent out to Ethan/laura to test tomorrow.  I will reintroduce the ones that make sense 🤙🏾

jcollins-axway (approved):

vivekschauhan (approved):

PR #570, opened 2022-10-17
APIGOV-23900 - remove wait groups that are not needed at the moment

jcollins-axway:


sbolosan (approved):

sbolosan:
i don't mind losing the logic... maybe just commenting out for now then?  i'm good either way.

jcollins-axway:
> i don't mind losing the logic... maybe just commenting out for now then? i'm good either way.

so this is the panic error, I think this needs to be better handled and the wait groups are not the solution

sbolosan:
make sense...  all good

PR #588, opened 2022-11-14
APIGOV-24087 - update migration

sbolosan:
Change queries to get only the latest revision

### DIFF
--- a/pkg/apic/apiservice.go
+++ b/pkg/apic/apiservice.go
@@ -11,6 +11,7 @@ import (
 
 	coreapi "github.com/Axway/agent-sdk/pkg/api"
 	apiv1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
+	v1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
 	management "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/management/v1alpha1"
 	defs "github.com/Axway/agent-sdk/pkg/apic/definitions"
 	utilerrors "github.com/Axway/agent-sdk/pkg/util/errors"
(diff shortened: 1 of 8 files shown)


pkg/apic/apiservice.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	coreapi "github.com/Axway/agent-sdk/pkg/api"
 	apiv1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
+	v1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
```
jcollins-axway, line 14: same import twice here

pkg/migrate/apisimigration.go
```diff
@@ -53,45 +52,23 @@ func (m *APISIMigration) Migrate(ri *apiv1.ResourceInstance) (*apiv1.ResourceIns
...
-
-			revisionInstances, err := m.getInstances(r)
-
-			instancesLock.Lock()
-			defer instancesLock.Unlock()
-			instances = append(instances, revisionInstances...)
 
-			errCh <- err
-		}(rev)
+	if len(revisions) > 0 {
```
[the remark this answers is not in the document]
jcollins-axway: With the changes I am looking to make we will not be able to make the assumption that the instance has a reference to the newest (winning) revision.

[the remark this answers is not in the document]
jcollins-axway: didn't see where this was, should be fine

pkg/apic/apiservicerevision.go
```diff
@@ -224,9 +224,11 @@ func (c *ServiceClient) setRevisionAction(serviceBody *ServiceBody) error {
...
-			"sort":   "metadata.audit.createTimestamp,DESC",
-			"fields": "name",
+			"query":    "metadata.references.id==" + serviceBody.serviceContext.serviceID,
+			"sort":     "metadata.audit.createTimestamp,DESC",
+			"page":     strconv.Itoa(1),
+			"pageSize": strconv.Itoa(1),
+			"fields":   "id",
 		}
 
 		revisions, err := c.GetAPIServiceRevisions(queryParams, c.cfg.GetRevisionsURL(), serviceBody.Stage)
```
jcollins-axway, line 234: Below this the revisionCount is being calculated by counting the length of the revisions returned. If it is limited to 1 the count will never be anything else.

We may still need to return all results, not limiting the pages.

Or, we can cease to name the API Service Revisions and let API Server do the naming. Not sure if/how this will affect the Revision title, however.

sbolosan: ah, good catch, let me think about that

vivekschauhan (approved):

PR #593, opened 2022-11-29
APIGOV-23982 - Register agent resource on agent initialization

vivekschauhan:
- Updates to create agent resource based on the agent type if resource does not exists
- Derive the agent name based on the environment name suffixed with short name for agent type
- Changes to update x-agent-detail sub resource on Environment resource to indicate the specific agent type is enabled in the environment


sbolosan (dismissed):
Looks good... do we need any migration path?

jcollins-axway (dismissed):

jcollins-axway (dismissed):

vivekschauhan:
> 



> Looks good... do we need any migration path?

No migration needed, the agent initialization updates the environment resource if the x-agent-details sub resource is not present or the agent type enabled property is not present. For agent resource creation, the initialization creates the agent resource if agent resources is not already present.

sbolosan (approved):

PR #606, opened 2022-12-20
APIGOV-24414 - Updated harvester client to manage purged sequences 

vivekschauhan:
- Changes to process 410 response from harvester and use the sequence to watch for following events.
- Fix in watch manager to initialize with no sequence

### DIFF
--- a/pkg/agent/poller/poller.go
+++ b/pkg/agent/poller/poller.go
@@ -61,7 +61,7 @@ func (m *pollExecutor) RegisterWatch(eventChan chan *proto.Event, errChan chan e
 		return
 	}
 
-	if m.sequence.GetSequence() == 0 {
+	if m.sequence.GetSequence() < 0 {
 		m.onHarvesterErr()
 		go func() {
 			m.Stop()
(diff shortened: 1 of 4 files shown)


sbolosan (approved):

pkg/agent/poller/poller.go
```diff
@@ -61,7 +61,7 @@ func (m *pollExecutor) RegisterWatch(eventChan chan *proto.Event, errChan chan e
 		return
 	}
 
-	if m.sequence.GetSequence() == 0 {
+	if m.sequence.GetSequence() < 0 {
```
jcollins-axway, line 64: Is there a concern to starting without a persistent cache? Would the sequence ID be 0 in this case?

vivekschauhan: In both mode(gRPC or poll), if persistent cache is not enabled initCache is called(https://github.com/Axway/agent-sdk/blob/APIGOV-24414/pkg/agent/eventsync.go#L91) which places call to get events from harvester with sequence ID 0 (all events) which initializes the resource cache and also sets the current sequence to the cache(https://github.com/Axway/agent-sdk/blob/3493072c0edb217dec9834323aaf9655e63c74c1/pkg/agent/eventsync.go#L121).

So this should be ok

jcollins-axway (dismissed):

jcollins-axway (approved):

PR #618, opened 2023-02-12
APIGOV-24821 - bubble up error from publish api

sbolosan:
when payload is over 10mb, log publish service error and return it up stream

### DIFF
--- a/pkg/apic/service.go
+++ b/pkg/apic/service.go
@@ -50,8 +50,8 @@ func (c *ServiceClient) PublishService(serviceBody *ServiceBody) (*management.AP
 	// there is a current envoy restriction with the payload size (10mb). Quick check on the size
 	if binary.Size(serviceBody.SpecDefinition) >= tenMB {
 		// if greater than 10mb, return
-		logErr := fmt.Sprintf("service %s carries a payload greater than 10mb. Service not created.", serviceBody.APIName)
-		return nil, errors.New(logErr)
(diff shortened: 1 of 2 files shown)


pkg/apic/service.go
```diff
@@ -50,8 +50,8 @@ func (c *ServiceClient) PublishService(serviceBody *ServiceBody) (*management.AP
 	// there is a current envoy restriction with the payload size (10mb). Quick check on the size
 	if binary.Size(serviceBody.SpecDefinition) >= tenMB {
 		// if greater than 10mb, return
-		logErr := fmt.Sprintf("service %s carries a payload greater than 10mb. Service not created.", serviceBody.APIName)
-		return nil, errors.New(logErr)
+		log.Error("error processing service")
```
jcollins-axway, line 53: can you create the error that is returned then log it before returning it?

Also fmt.Errorf == errors.New(fmt.Sprintf(...))

```
err := fmt.Errorf("....", apiname)
logger.WithError(err).Error("error pro...")
return nil, err
```

sbolosan: @jcollins-axway  yes, but i logged a generic one and then returned the fmt.Sprintf one...  assuming if I log and return the same error, it prints 2x? or not... so, i thought, log generic, and then the more specific one is returned and logged as well.

jcollins-axway (approved):

PR #628, opened 2023-03-14
APIGOV-25078 - Changes to reduce the status updates on startup healthcheck failures

vivekschauhan:
- Updates to register status update jobs after the startup healthcheck to reduce status update calls due to restarts of the failing agent running as service
- Changes to use the existing subjects on the ACL resource to verify if the ACL needs to be updated.
- Changes to use PUT call for updating the ACL resource instead of DELETE followed by POST

### DIFF
--- a/pkg/agent/aclupdatejob_test.go
+++ b/pkg/agent/aclupdatejob_test.go
@@ -126,8 +126,8 @@ func TestACLUpdateHandlerJob(t *testing.T) {
 				if strings.Contains(req.RequestURI, "/apis/management/v1alpha1/environments/"+test.envName+"/accesscontrollists") {
 					aclReturn, _ := io.ReadAll(req.Body)
 					switch {
-					case req.Method == http.MethodDelete:
-						resp.WriteHeader(http.StatusNoContent)
+					case req.Method == http.MethodPut:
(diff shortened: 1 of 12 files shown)


sbolosan (approved):

pkg/agent/aclupdatejob_test.go
```diff
@@ -126,8 +126,8 @@ func TestACLUpdateHandlerJob(t *testing.T) {
 				if strings.Contains(req.RequestURI, "/apis/management/v1alpha1/environments/"+test.envName+"/accesscontrollists") {
 					aclReturn, _ := io.ReadAll(req.Body)
 					switch {
-					case req.Method == http.MethodDelete:
-						resp.WriteHeader(http.StatusNoContent)
+					case req.Method == http.MethodPut:
+						fallthrough
 					case req.Method == http.MethodPost:
```
jcollins-axway, line 131: This switch is not needed any longer, response is always the same

jcollins-axway (approved):

pkg/agent/aclupdatejob_test.go
vivekschauhan: Right, fixed

PR #629, opened 2023-03-15
APIGOV-25071 - handle multiple categories with same title

jcollins-axway:


pkg/agent/cache/category.go
```diff
@@ -13,7 +14,20 @@ import (
 func (c *cacheManager) AddCategory(resource *v1.ResourceInstance) {
 	defer c.setCacheUpdated(true)
 
-	c.categoryMap.SetWithSecondaryKey(resource.Name, fmt.Sprintf("title-%s", resource.Title), resource)
+	// get any existing categories with the same title, the oldest one should win the secondary key
+	if cachedCategory := c.GetCategoryWithTitle(resource.Title); cachedCategory != nil {
```
sbolosan, line 18: this really only returns 1 title (from the map), right?

jcollins-axway: Secondary Keys are unique in the cache

sbolosan (approved):

vivekschauhan (approved):

PR #644, opened 2023-05-18
APIGOV-25482 - move credential expiry to environment level

sbolosan:


vivekschauhan (commented):
Current change will address setting the credential expiration config on agent startup, however we will need to update the watch topic as well to include update events on Environment resource and update the credential expiration config on receiving update event.

pkg/agent/events/watchtopic.go
```diff
@@ -295,6 +297,7 @@ func NewGovernanceAgentWatchTopic(name, scope string, agentResourceGroupKind v1.
 			{GroupKind: management.ManagedApplicationGVK().GroupKind, ScopeName: scope, EventTypes: createdOrUpdated},
 			{GroupKind: management.CredentialRequestDefinitionGVK().GroupKind, ScopeName: scope, EventTypes: all},
 			{GroupKind: management.AccessRequestDefinitionGVK().GroupKind, ScopeName: scope, EventTypes: all},
+			{GroupKind: management.EnvironmentGVK().GroupKind, Name: scope, EventTypes: updated},
```
vivekschauhan, line 300: We don't need updates on governance agent watch topic

vivekschauhan (approved):

pkg/agent/handler/environment.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+func (c *environmentHandler) Handle(ctx context.Context, meta *proto.EventMeta, resource *v1.ResourceInstance) error {
+	action := GetActionFromContext(ctx)
+	if resource.Kind != environment.EnvironmentGVK().Kind {
+		return nil
+	}
+
+	if action == proto.Event_SUBRESOURCEUPDATED && meta.Subresource == management.EnvironmentPoliciesSubResourceName {
```
jcollins-axway, line 33: Can you change this to return nil if action is not subresource or subresource not policies? That way the meat of the function is not within the fi statement?

jcollins-axway:
I also saw @vivekschauhan comment on not needed to update the governance agent, can't we technically remove that altogether as the governance agent no longer exists as a resource?

jcollins-axway:
> I also saw @vivekschauhan comment on not needed to update the governance agent, can't we technically remove that altogether as the governance agent no longer exists as a resource?

This not the story for that though

jcollins-axway (approved):

vivekschauhan (approved):

vivekschauhan:
> > I also saw @vivekschauhan comment on not needed to update the governance agent, can't we technically remove that altogether as the governance agent no longer exists as a resource?
> 
> This not the story for that though

Probably should still keep it

jcollins-axway:
> > > I also saw @vivekschauhan comment on not needed to update the governance agent, can't we technically remove that altogether as the governance agent no longer exists as a resource?
> > 
> > 
> > This not the story for that though
> 
> Probably should still keep it

Why would we keep this here? When we generated the resources at one point I had to remove the governance agent references as they no longer exist.

PR #654, opened 2023-06-09
APIGOV- 25443 Agent SDK: Update event builders and event definitions to support SaaS agent

kasuvy:


jcollins-axway:
Looks good @kasuvy. We can look at adding that test next.

sbolosan (approved):

pkg/transaction/httpprotocolbuilder.go
```diff
@@ -375,9 +385,14 @@ func (b *httpProtocolBuilder) queryArgsRedaction() {
 	if b.err != nil {
 		return
 	}
-
+	var redactedArgs map[string][]string
```
vivekschauhan, line 388: We should move setting the p.httpProtocol.URI in SetURI as well here under the Build to use the overridden redactionConfig instead.

kasuvy: Fixed in the latest commit.

jcollins-axway (approved):

PR #655, opened 2023-06-15
APIGOV-25443 Added support for setting redaction config on logEventBuilder

kasuvy:


pkg/transaction/definitions.go
```diff
@@ -50,20 +50,22 @@ const (
...
+	EnvironmentName        string   `json:"environmentName,omitempty"`
+	EnvironmentID          string   `json:"environmentId"`
+	TenantID               string   `json:"tenantId"`
+	TrcbltPartitionID      string   `json:"trcbltPartitionId"`
+	Type                   string   `json:"type"`
+	TargetPath             string   `json:"targetPath,omitempty"`
+	ResourcePath           string   `json:"resourcePath,omitempty"`
+	TransactionEvent       *Event   `json:"transactionEvent,omitempty"`
+	TransactionSummary     *Summary `json:"transactionSummary,omitempty"`
+	UnRedactedTargetPath   string
```
vivekschauhan, line 67: Should add json tag((`json:"-"`) to not serialize these properties

kasuvy: Fixed it in another commit.

pkg/transaction/logeventbuilder.go
```diff
@@ -300,11 +304,45 @@ func (b *transactionEventBuilder) SetProtocolDetail(protocolDetail interface{})
...
+	} else {
+		b.logEvent.TargetPath, b.err = b.redactionConfig.URIRedaction(b.logEvent.UnRedactedTargetPath)
+	}
+	if b.err != nil {
+		return nil, b.err
+	}
+
+	//Set Resource Path
+	if b.logEvent.UnRedactedResourcePath == "" {
+		return nil, errors.New("unredacted resource path property not set in the LogEvent details")
```
jcollins-axway, line 335: what if b.logEvent.UnRedactedResourcePath is set while b.logEvent.UnRedactedTargetPath is not, looks like b.logEvent.UnRedactedResourcePath would not be redacted

pkg/transaction/definitions.go
```diff
@@ -113,10 +113,11 @@ type Runtime struct {
...
 type EntryPoint struct {
-	Type   string `json:"type,omitempty"`
-	Method string `json:"method,omitempty"`
-	Path   string `json:"path,omitempty"`
-	Host   string `json:"host,omitempty"`
+	Type           string `json:"type,omitempty"`
+	Method         string `json:"method,omitempty"`
+	Path           string `json:"path,omitempty"`
+	Host           string `json:"host,omitempty"`
+	UnRedactedPath string `json:"-"`
```
jcollins-axway, line 120: Could you make the same change we did for the targetpath and resourcepath here, removing the UnRedacted* vars?

jcollins-axway (changes requested):
comment above

pkg/transaction/definitions.go
kasuvy: Added the change.

jcollins-axway (approved):

sbolosan (approved):

PR #663, opened 2023-07-12
SOL-12 access request and credential request onDeleting managed app

mynecker:
added managed app to onDeleting for access request and credential request


pkg/agent/handler/accessrequest.go
```diff
@@ -186,10 +186,17 @@ func (h *accessRequestHandler) onError(_ context.Context, ar *management.AccessR
 
 // onDeleting deprovisions an access request and removes the finalizer
 func (h *accessRequestHandler) onDeleting(ctx context.Context, ar *management.AccessRequest) {
+	log := getLoggerFromContext(ctx)
+
+	app, err := h.getManagedApp(ctx, ar)
+	if err != nil {
+		log.WithError(err).Error("error getting managed app")
+		h.onError(ctx, ar, err)
```
jcollins-axway, line 194: @mynecker Could we return the err here if one is hit? All else looks good, so after this change it will be merged and a release made soon.

mynecker: @jcollins-axway 

I can just 'return' if we hit an error, is that sufficient? 

```
	app, err := h.getManagedApp(ctx, ar)
	if err != nil {
		log.WithError(err).Error("error getting managed app")
		h.onError(ctx, ar, err)
		return
	}
```

jcollins-axway: yes, sorry, you are correct that this method does not return a type. The onError will update the resource with the error

jcollins-axway: @mynecker if you can make this change we can get this merged ahead of the next release of the sdk

mynecker: @jcollins-axway should handle errors now as expected

jcollins-axway: Thanks!

jcollins-axway (approved):

sbolosan (approved):
solid...

vivekschauhan (approved):

PR #672, opened 2023-08-08
APIGOV-24312 - Support for using additional auth when communicating to IdP

vivekschauhan:
- Changes to configure the agents to use client_secret_basic, client_secret_post, client_secret_jwt, private_key_jwt, tls_client_auth and self_signed_tls_client_auth
- Changes to acquire token based for the registered IdP provider on the configured auth types


pkg/authz/oauth/constants.go
```diff
@@ -16,8 +16,14 @@ const (
 
 // IDP Auth type string const
 const (
-	IDPAuthTypeAccessToken = "accessToken"
-	IDPAuthTypeClient      = "client"
+	IDPAuthTypeAccessToken             = "accessToken"
+	IDPAuthTypeClient                  = "client"
+	IDPAuthTypeClientSecretBasic       = "client_secret_basic"
```
jcollins-axway, line 21: this must be the default from the IDP but camelCase vs underscore_seperation bothers me a little

pkg/config/externalidpconfig.go
```diff
@@ -11,6 +11,12 @@ import (
 const (
 	accessToken              = "accessToken"
 	client                   = "client"
+	clientSecretBasic        = "client_secret_basic"
+	clientSecretPost         = "client_secret_post"
+	clientSecretJwt          = "client_secret_jwt"
+	privateKeyJwt            = "private_key_jwt"
+	tlsClientAuth            = "tls_client_auth"
+	selfSignedTlsClientAuth  = "self_signed_tls_client_auth"
```
jcollins-axway, line 19: do these constants and the other constants have to be defined separately?

jcollins-axway (dismissed):

pkg/config/externalidpconfig.go
vivekschauhan: These constants are based on the token authentication types coming from OAuth specification

vivekschauhan: Will use these and remove the ones in oauth package

PR #684, opened 2023-09-18
APIGOV-26331 - check to make sure client is not nil

sbolosan:
When in offline mode, the client isn't being set (and doesn't need to be).  Check for nil ...

### DIFF
--- a/pkg/transaction/eventgenerator.go
+++ b/pkg/transaction/eventgenerator.go
@@ -148,12 +148,10 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
 	}
 
 	// Check to see if marketplace provisioning/subs is enabled
-	if agent.GetCentralClient() != nil {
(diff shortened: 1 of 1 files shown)


pkg/transaction/eventgenerator.go
```diff
@@ -148,10 +148,12 @@ func (e *Generator) CreateEvents(summaryEvent LogEvent, detailEvents []LogEvent,
 	}
 
 	// Check to see if marketplace provisioning/subs is enabled
-	if agent.GetCentralClient().IsMarketplaceSubsEnabled() {
-		err := e.processTxnSummary(summaryEvent)
-		if err != nil {
-			return nil, err
+	if agent.GetCentralClient() != nil {
+		if agent.GetCentralClient().IsMarketplaceSubsEnabled() {
```
jcollins-axway, line 152: this can be simplified

if agent.GetCentralClient() != nil && agent.GetCentralClient().IsMarketplaceSubsEnabled() {...}

jcollins-axway (approved):

PR #688, opened 2023-09-28
APIGOV-25807 github dataplane settings

alrosca:


pkg/apic/apiserver/models/management/v1alpha1/model_dataplane_spec_git_hub.go
```diff
@@ -0,0 +1,18 @@
...
+ * Generated by: OpenAPI Generator (https://openapi-generator.tech)
+ */
+
+package management
+
+// DataplaneSpecGitHub The configuration common to all GitHub agents that use this dataplane
+type DataplaneSpecGitHub struct {
+	Type            string                            `json:"type,omitempty"`
+	RepositoryName  string                            `json:"repositoryName"`
+	RepositoryOwner string                            `json:"repositoryOwner"`
```
jcollins-axway, line 16: I am not a fan of repeating the word repository for both these properties. Can we drop that in api-server and just make them name and owner?

pkg/apic/apiserver/models/management/v1alpha1/model_dataplane_spec_git_hub.go
```diff
@@ -0,0 +1,18 @@
...
+	FilenameFilter  DataplaneSpecGitHubFilenameFilter `json:"filenameFilter,omitempty"`
```
jcollins-axway, line 17: With this object now having a path, can this property just be Filter?

PR #697, opened 2023-11-06
APIGOV-26631 - set up crd on startup

sbolosan:


pkg/agent/agent.go
```diff
@@ -257,6 +269,51 @@ func registerCredentialProvider(idp config.IDPConfig, tlsCfg config.TLSConfig, p
 			WithField("metadata-url", idp.GetMetadataURL()).
 			Errorf("unable to register external IdP provider, any credential request to the IdP will not be processed. %s", err.Error())
 	}
+	crdName := idp.GetIDPTitle() + " " + provisioning.OAuthIDPCRD
```
vivekschauhan, line 272: This should be idp.GetIDPName(), title is a friendly name. If title is not set by the config, it gets set as name during the config initalization/validation

sbolosan: ah, yes...  you're right.  Not sure why I put title there.

vivekschauhan (approved):

jcollins-axway (approved):

PR #709, opened 2023-12-13
APIGOV-26666 - change api validator frequency to cron schedule

dgghinea:


sbolosan (dismissed):
Looks good. We will need to update docs

pkg/config/centralconfig.go
```diff
@@ -783,7 +784,7 @@ func AddCentralConfigProperties(props properties.Properties, agentType AgentType
 	props.AddStringProperty(pathProxyURL, "", "The Proxy URL to use for communication to Amplify Central")
 	props.AddDurationProperty(pathPollInterval, 60*time.Second, "The time interval at which the central will be polled for subscription processing")
 	props.AddDurationProperty(pathReportActivityFrequency, 5*time.Minute, "The time interval at which the agent polls for event changes for the periodic agent status updater")
-	props.AddDurationProperty(pathAPIValidationFrequency, 1*time.Hour, "The time interval at which the agent validates API Services with the dataplane", properties.WithLowerLimit(5*time.Minute), properties.WithQAOverride())
+	props.AddStringProperty(pathAPIValidationCronSchedule, "0 * * * *", "The cron schedule at which the agent validates API Services with the dataplane")
```
vivekschauhan, line 787: Too aggressive schedule, I guess we decided to use daily as default

vivekschauhan (approved):

alrosca (approved):

pkg/config/centralconfig.go
```diff
@@ -783,7 +784,7 @@ func AddCentralConfigProperties(props properties.Properties, agentType AgentType
 	props.AddStringProperty(pathProxyURL, "", "The Proxy URL to use for communication to Amplify Central")
 	props.AddDurationProperty(pathPollInterval, 60*time.Second, "The time interval at which the central will be polled for subscription processing")
 	props.AddDurationProperty(pathReportActivityFrequency, 5*time.Minute, "The time interval at which the agent polls for event changes for the periodic agent status updater")
-	props.AddDurationProperty(pathAPIValidationFrequency, 1*time.Hour, "The time interval at which the agent validates API Services with the dataplane", properties.WithLowerLimit(5*time.Minute), properties.WithQAOverride())
+	props.AddStringProperty(pathAPIValidationCronSchedule, "0 0 * * *", "The cron schedule at which the agent validates API Services with the dataplane")
```
jcollins-axway, line 787: the default can simply be @daily which is equivalent to 0 0 * * *, its just easier to understand at a glance

dgghinea: oh. I actually didn't know you could do that. thanks!

jcollins-axway: We probably need to add a QA override and do some cron checks to make sure it is not configured for more often than our previous lower limit of 5 minutes.

Check out the cron schedule validation in the usage config, we essentially get the next number of x cron scheduled intervals and check that the difference between them is never lower than our limit.

vivekschauhan (approved):

jcollins-axway (approved):

PR #711, opened 2023-12-15
APIGOV-26213 - Issue in adding scopes in marketplace client creation page

dgghinea:
marketplace scopes fix (hopefully)

### DIFF
--- a/pkg/apic/provisioning/propertybuilder.go
+++ b/pkg/apic/provisioning/propertybuilder.go
@@ -44,6 +44,7 @@ type propertyDefinition struct {
 	IsEncrypted        bool                          `json:"x-axway-encrypted,omitempty"`
 	Widget             string                        `json:"x-axway-widget,omitempty"`
 	IsCopyable         bool                          `json:"x-axway-copyable,omitempty"`
+	ScopesUniqueItems  bool                          `json:"uniqueItems,omitempty"`
 	Name               string                        `json:"-"`
 	Required           bool                          `json:"-"`
 }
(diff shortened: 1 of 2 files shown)


pkg/apic/provisioning/propertybuilder.go
```diff
@@ -44,6 +44,7 @@ type propertyDefinition struct {
 	IsEncrypted        bool                          `json:"x-axway-encrypted,omitempty"`
 	Widget             string                        `json:"x-axway-widget,omitempty"`
 	IsCopyable         bool                          `json:"x-axway-copyable,omitempty"`
+	ScopesUniqueItems  bool                          `json:"uniqueItems,omitempty"`
```
jcollins-axway, line 47: the name of the attribute is misleading, as this attribute should be available for any ArrayProperty.

perhaps a new builder method needs added on that struct?

pkg/apic/provisioning/propertybuilder.go
```diff
@@ -187,6 +189,9 @@ func (p *schemaProperty) SetName(name string) TypePropertyBuilder {
 // SetLabel - sets the label of the property
 func (p *schemaProperty) SetLabel(label string) TypePropertyBuilder {
 	p.label = label
+	if p.label == OauthScopes {
+		p.uniqueItems = true
+	}
```
jcollins-axway, line 194: this is too specific on how it expects the oath scopes, as mentioned above setting unique property should be controlled by the implementer of the builder and not limited to the label placed on the property

jcollins-axway (approved):

PR #712, opened 2023-12-20
APIGOV-26877 - update usage report logic

dgghinea:


pkg/transaction/metric/metricspublisher.go
```diff
@@ -67,13 +68,30 @@ func (c *metricPublisher) publishToLighthouse(event LighthouseUsageEvent) error
...
+	if response.Code < 400 {
+		return nil
+	} else if response.Code >= 500 {
+		err := fmt.Errorf("Server error")
+		c.logger.WithField("statusCode", response.Code).WithError(err).Error(string(response.Body))
+		return err
+	}
+
+	usageResp := LighthouseUsageResponse{}
+	unmarshalErr := json.Unmarshal(response.Body, &usageResp)
```
jcollins-axway, line 81: We usually just name this err rather than a specific named err.

pkg/transaction/metric/metricspublisher.go
```diff
@@ -67,13 +68,30 @@ func (c *metricPublisher) publishToLighthouse(event LighthouseUsageEvent) error
...
+	if unmarshalErr != nil {
+		c.logger.WithField("responseBody", string(response.Body)).WithField("statusCode", response.Code).Error(unmarshalErr)
```
jcollins-axway, line 83: update this to 

c.logger.WithField...WithField...WithError(unmarshalErr).Error("helpful message")

pkg/transaction/metric/metricspublisher.go
```diff
@@ -67,13 +68,30 @@ func (c *metricPublisher) publishToLighthouse(event LighthouseUsageEvent) error
...
+		return unmarshalErr
+	}
+	if strings.HasPrefix(usageResp.Description, "The file exceeds the maximum upload size of ") || usageResp.Description == "Environment ID not found" {
+		err := fmt.Errorf("request failed with unexpected status code. Not scheduling for retry in the next batch.")
+		c.logger.WithField("statusCode", response.Code).WithError(err).Error(usageResp.Description)
+		return nil
+	} else {
```
jcollins-axway, line 90: This else looks to be unnecessary since the if block returns.

pkg/transaction/metric/metricspublisher.go
```diff
@@ -67,13 +68,30 @@ func (c *metricPublisher) publishToLighthouse(event LighthouseUsageEvent) error
 		c.logger.WithError(err).Error("publishing usage")
 		return err
 	}
-	if response.Code >= 400 {
-		resBody := string(response.Body)
+
+	if response.Code < 400 {
+		return nil
```
jcollins-axway, line 73: Should this look specifically for the 202 response code and not just under 400? That way we get some sort of log when its an unexpected response, likely in the unmarshal error block.

jcollins-axway (approved):

PR #716, opened 2024-01-09
APIGOV-26893 - new usage report generation logic

dgghinea:


jcollins-axway:
This seems to update the report as it is going into the cache, which is not the desired place for this. We want the aggregation to occur as the report is being sent to the platform and to not affect the offline reporting that this change looks like it will.

jcollins-axway (changes requested):
see above comment

jcollins-axway (approved):

vivekschauhan (approved):

PR #724, opened 2024-01-23
APIGOV-27081 - update elastic/beats

dfeldick:


jcollins-axway:
not that I think it'll cause issues, but should we wait to merge this until after we release?

PR #735, opened 2024-02-15
APIGOV-27238 - fix single usage report granularity

dgghinea:


pkg/transaction/metric/metricspublisher.go
```diff
@@ -138,17 +135,13 @@ func aggregateReports(event LighthouseUsageEvent) LighthouseUsageEvent {
...
-				finalReport[orderedKeys[0]].Usage[usageKey] = usageVal
-			}
+			finalReport[orderedKeys[0]].Usage[usageKey] += usageVal
 		}
 	}
 	event.Report = finalReport
 
 	startTime, _ := time.Parse(ISO8601, orderedKeys[0])
-	endTime, _ := time.Parse(ISO8601, orderedKeys[len(orderedKeys)-1])
+	endTime := now()
```
jcollins-axway, line 144: Could you update the event.Timestamp with endTime as well?

jcollins-axway:
Thinking about this issue a bit today, want @vivekschauhan's opinion too.

When the actual report sending is done we are sending from the time of the first report to the end time of the last report. Not necessarily the time of the first report to the time the send is happening.

With that said...should we additionally execute the metric collection job prior to this send? That way the usage is up to date to the execution time of the report sending?

The gap could be anywhere for 0-15 minutes if we do not do this.

jcollins-axway:
> Thinking about this issue a bit today, want @vivekschauhan's opinion too.
> 
> When the actual report sending is done we are sending from the time of the first report to the end time of the last report. Not necessarily the time of the first report to the time the send is happening.
> 
> With that said...should we additionally execute the metric collection job prior to this send? That way the usage is up to date to the execution time of the report sending?
> 
> The gap could be anywhere for 0-15 minutes if we do not do this.

@vivekschauhan agreed, we should call execute on the collector prior to publishing the metrics.

jcollins-axway (dismissed):

jcollins-axway (approved):

PR #747, opened 2024-03-18
APIGOV-EmbeddedFix - filter out _embedded

sbolosan:
add _embedded to the filter fields.


sbolosan (commented):
Will wait for PR approval and @deepak confirmation

jcollins-axway (approved):

PR #771, opened 2024-05-09
APIGOV-27848 - usage logger

dgghinea:


pkg/config/logconfig.go
```diff
@@ -16,17 +16,19 @@ import (
...
-	Output       string               `config:"output"`
-	File         LogFileConfiguration `config:"file"`
-	MetricFile   LogFileConfiguration `config:"metricfile"`
-	MaskedValues string               `config:"maskedvalues"`
+	Level                     string               `config:"level"`
+	Format                    string               `config:"format"`
+	Output                    string               `config:"output"`
+	File                      LogFileConfiguration `config:"file"`
+	MetricFile                LogFileConfiguration `config:"metricfile"`
+	PublishedTransactionsFile LogFileConfiguration `config:"publishedtransactionsfile"`
```
vivekschauhan, line 30: Can we  usageFile (or something similar) to not confuse usage audit with transactions

dgghinea: sure, will do.

pkg/config/logconfig.go
```diff
@@ -43,13 +45,20 @@ func (l *LogConfiguration) setupLogger(agentType AgentType) error {
 		MaxMetricSize(l.MetricFile.MaxSize).
 		MaxMetricBackups(l.MetricFile.MaxBackups).
 		MaxMetricAge(l.MetricFile.MaxAge).
+		Usage(agentType == TraceabilityAgent && l.UsageFile.Enabled).
+		UsageFilename(l.UsageFile.Name).
+		MaxPublishedTransactionAge(l.UsageFile.MaxAge).
```
jcollins-axway, line 50: I think we may need to add all the rotate properties here. The way the options here are used will not map specifically to how the story lays it out. But as long as we allow them to tune the settings they can do what they want.

PR #818, opened 2024-08-08
APIGOV-DG000 - fix ARD creation

dgghinea:
Adds a flag that ignores creating ARD and CRDs based on the spec provided in the serviceBody

### DIFF
--- a/pkg/apic/servicebody.go
+++ b/pkg/apic/servicebody.go
@@ -96,7 +96,7 @@ func (s *ServiceBody) GetScopes() map[string]string {
 
 // GetCredentialRequestDefinitions - returns the array of all credential request policies
 func (s *ServiceBody) GetCredentialRequestDefinitions() []string {
-	if len(s.credentialRequestPolicies) > 0 {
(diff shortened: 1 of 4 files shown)


jcollins-axway (dismissed):

pkg/apic/servicebody.go
```diff
@@ -117,6 +117,9 @@ func (s *ServiceBody) GetAccessRequestDefinition() *management.AccessRequestDefi
 }
 
 func (s *ServiceBody) createAccessRequestDefinition() error {
+	if !s.uniqueARD {
+		return nil
```
sbolosan, line 121: can you log s.ardName  before you return since we are bailing right out

jcollins-axway: Thinking debug level? Since this would happen a lot I think Warn and Info would not apply. More so since this only happens when the user said the ard is not unique for this api.

sbolosan (dismissed):

jcollins-axway (dismissed):

jcollins-axway (approved):

PR #823, opened 2024-08-22
APIGOV-28627 - bubble any error when discovering spec type

sbolosan:


jcollins-axway:
Was there a decision about not logging the error at all? In the ticket I think the suggestion was to log something in the oas spec parsing if error occurs there.

sbolosan:
It bubbles up to where .Parse() is being called in the body builder... and that returns to the agents

	specParser := NewSpecResourceParser(b.serviceBody.SpecDefinition, b.serviceBody.ResourceType)
	err := specParser.Parse()
	if err != nil {
		return b.serviceBody, fmt.Errorf("failed to parse service specification for '%s': %s", b.serviceBody.APIName, err)
	}

 and in the agent, I could see that it was logging the error

maybe I missed it, but I did not see any other flow where explicitly oas spec parsing was calling .Parse()

pkg/util/oas/oas.go
```diff
@@ -16,7 +16,7 @@ func ParseOAS2(spec []byte) (*openapi2.T, error) {
 	swaggerObj := &openapi2.T{}
 	err := json.Unmarshal(spec, swaggerObj)
 	if err != nil {
-		return nil, err
+		return nil, errors.New("unable to parse OAS2 specification")
```
jcollins-axway, line 19: I think we should just log an error here, with the details, and return as is. Same with the other one.

jcollins-axway (approved):

vivekschauhan (approved):

PR #826, opened 2024-09-04
APIGOV-28410 usage reports on schedule

alrosca:


jcollins-axway (dismissed):

pkg/transaction/metric/usagepublisher.go
```diff
@@ -233,8 +235,11 @@ func (c *usagePublisher) Ready() bool {
 
 // Execute - process the offline report generation
 func (c *usagePublisher) Execute() error {
-	if c.offline {
-		return c.report.saveReport()
+	if c.report.shouldPublish(c.schedule) {
```
vivekschauhan, line 238: Should we move down this check to limit only sendReport and not to saveReport?

jcollins-axway: I think it could make sense how it is, as then offline reports will not be updated any extra either. Thoughts?

PR #842, opened 2024-10-08
APIGOV-28932 - remove jfrog call and use agentstate

dgghinea:
- Removes jfrog call and uses the agentstate from the DA/TA resource(which are updated by separate jobs(for on-prem) and agents-controller(for SaaS)
 - in case the state is not found, we simply log a warning with it(this can happen the first time the agent starts)

### DIFF
--- a/pkg/cmd/agentversionjob.go
+++ b/pkg/cmd/agentversionjob.go
@@ -9,9 +9,6 @@ import (
 	"github.com/Axway/agent-sdk/pkg/config"
 	"github.com/Axway/agent-sdk/pkg/util"
 
-	"regexp"
-	"strings"
-
 	"github.com/Axway/agent-sdk/pkg/jobs"
 	"github.com/Axway/agent-sdk/pkg/util/errors"
 	log "github.com/Axway/agent-sdk/pkg/util/log"
@@ -24,26 +21,22 @@ const (
 // AgentVersionCheckJob - polls for agent versions
 type AgentVersionCheckJob struct {
 	jobs.Job
-	buildVersion string
-	manager      resource.Manager
+	logger  log.FieldLogger
+	manager resource.Manager
 }
 
 // NewAgentVersionCheckJob - creates a new agent version check job structure
 func NewAgentVersionCheckJob(cfg config.CentralConfig) (*AgentVersionCheckJob, error) {
-	// get current build version
-	buildVersion, err := getBuildVersion()
-	if err != nil {
-		return nil, err
-	}
-
 	manager := agent.GetAgentResourceManager()
 	if manager == nil {
 		return nil, errors.ErrStartingVersionChecker.FormatError("could not get the agent resource manager")
 	}
 
 	return &AgentVersionCheckJob{
-		manager:      manager,
-		buildVersion: buildVersion,
+		manager: manager,
+		logger: log.NewFieldLogger().
+			WithPackage("sdk.cmd").
+			WithComponent("agentVersionJob"),
 	}, nil
 }
 
@@ -61,14 +54,15 @@ func (avj *AgentVersionCheckJob) Status() error {
 func (avj *AgentVersionCheckJob) Execute() error {
 	state, err := avj.getAgentState()
 	if err != nil {
-		log.Trace(err)
-		// Could not get update from agent state.  Warn that we could not determine version and continue processing
-		log.Warn("Agent cannot determine the current available release. Be aware that your agent could be outdated.")
+		avj.logger.WithError(err).Warn("agent cannot determine the current available release. Be aware that your agent could be outdated.")
+		return nil
 	}
 
 	switch state {
+	case "current":
+		log.Trace("agent is up to date.")
 	case "available":
-		log.Warn("Please be aware that there is a newer agent version available")
+		log.Warn("please be aware that there is a newer agent version available.")
 	case "outdated":
 		log.Error("current agent version is no longer supported. We strongly advise to update the agent as soon as possible.")
 	case "retracted":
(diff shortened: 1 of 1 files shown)


pkg/cmd/agentversionjob.go
```diff
@@ -128,44 +59,51 @@ func (avj *AgentVersionCheckJob) Status() error {
...
+		return "", fmt.Errorf("could not get the agent resource")
 	}
-	response, err := avj.apiClient.Send(request)
-	if err != nil {
-		return err
+
+	// The kind should be only DA or TA
+	subResKey := management.DiscoveryAgentAgentstateSubResourceName
+	if agentRes.GetGroupVersionKind().Kind == "TraceabilityAgent" {
+		subResKey = management.DiscoveryAgentAgentstateSubResourceName
```
vivekschauhan, line 89: Although it the same value but should be TraceabilityAgentAgentstateSubResourceName

pkg/cmd/agentversionjob.go
```diff
@@ -128,44 +59,51 @@ func (avj *AgentVersionCheckJob) Status() error {
...
 
-	jfrogResponse := jfrogData{}
-	err = json.Unmarshal(response.Body, &jfrogResponse)
-	if err != nil {
-		return err
+	err := fmt.Errorf("could not find the agentstate from agent subresource")
+	// This can happen at the first time the job executes, in which the resource has no agentState set beforehand
+	agentStateIface := agentRes.GetSubResource(subResKey)
+	if agentStateIface == nil {
+		return "", err
```
vivekschauhan, line 96: We probably should ignore error if there is no agentState for existing agent as they will not have agentState set until the agent resource is updated.

vivekschauhan (dismissed):

pkg/cmd/agentversionjob.go
dgghinea: the code that calls this function doesn't pass the error, only logs it. that was my thought behind it. but I can change this as well.

pkg/cmd/agentversionjob.go
```diff
@@ -1,118 +1,49 @@
...
-	Local        bool   `json:"local,omitempty"`
-	Type         string `json:"type,omitempty"`
-	Compacted    bool   `json:"compacted,omitempty"`
-	Cached       bool   `json:"cached,omitempty"`
-	Trash        bool   `json:"trash,omitempty"`
-	Distribution bool   `json:"distribution,omitempty"`
-}
-
 // AgentVersionCheckJob - polls for agent versions
 type AgentVersionCheckJob struct {
```
jcollins-axway, line 25: lets add a field logger here like we do for other packages and use it through out this job

pkg/cmd/agentversionjob.go
```diff
@@ -128,44 +59,51 @@ func (avj *AgentVersionCheckJob) Status() error {
...
 		log.Trace(err)
-		// Could not get update from jfrog.  Warn that we could not determine version and continue processing
-		log.Warn("Agent cannot determine the next available release. Be aware that your agent could be outdated.")
-	} else {
-		// Successfully got jfrog version.  Now compare build to latest version
-		if isVersionStringOlder(avj.buildVersion, config.AgentLatestVersion) {
-			log.Warnf("New version available. Please consider upgrading from version %s to version %s", avj.buildVersion, config.AgentLatestVersion)
-		}
+		// Could not get update from agent state.  Warn that we could not determine version and continue processing
+		log.Warn("Agent cannot determine the current available release. Be aware that your agent could be outdated.")
```
jcollins-axway, line 66: maybe we should return after adding this warning

jcollins-axway: I guess its 6 in one half dozen in the other but does avoid the switch statement

pkg/cmd/agentversionjob.go
```diff
@@ -128,44 +59,51 @@ func (avj *AgentVersionCheckJob) Status() error {
...
+	}
+
+	switch state {
+	case "available":
+		log.Warn("Please be aware that there is a newer agent version available")
+	case "outdated":
+		log.Error("current agent version is no longer supported. We strongly advise to update the agent as soon as possible.")
+	case "retracted":
+		log.Error("current agent version has a known issue, please update the agent immediately.")
```
jcollins-axway, line 75: should we consider adding a current case and logging Trace level?

pkg/cmd/agentversionjob.go
```diff
@@ -128,44 +59,51 @@ func (avj *AgentVersionCheckJob) Status() error {
...
+		return "", fmt.Errorf("could not get the agent resource")
 	}
-	response, err := avj.apiClient.Send(request)
-	if err != nil {
-		return err
+
+	// The kind should be only DA or TA
+	subResKey := management.DiscoveryAgentAgentstateSubResourceName
+	if agentRes.GetGroupVersionKind().Kind == "TraceabilityAgent" {
+		subResKey = management.TraceabilityAgentAgentstateSubResourceName
```
jcollins-axway, line 89: Technically the subreskey will be the same, but maybe we determine the kind and then convert the agentRes instance to the appropriate kind. Then further down we can use the .Agentstate.update dot notation?

pkg/cmd/agentversionjob.go
sbolosan: Can I get a confirmation on case for logging at the start of the of the sentence? And does it matter what level? We have both upper and lower case here

jcollins-axway:
digging all the cleaned code

pkg/cmd/agentversionjob.go
```diff
@@ -128,129 +52,47 @@ func (avj *AgentVersionCheckJob) Status() error {
...
-// In the future, adding a (Generic) resource for grouping versions together under the same scope is a possible solution
-// ie: a new unscoped resource that represents the platform services, so that other products can plug in their releases.
-func (avj *AgentVersionCheckJob) getJFrogVersions() error {
-	request := coreapi.Request{
-		Method:  http.MethodPost,
-		URL:     jfrogURL,
-		Headers: avj.headers,
-		Body:    avj.requestBytes,
+	switch state {
+	case "current":
```
vivekschauhan, line 62: Can we have constants for the states?

vivekschauhan (dismissed):

alrosca (approved):

jcollins-axway:
@dgghinea should this have been merged?

dgghinea:
> @dgghinea should this have been merged?

no one did a quick test with this, afaik

PR #844, opened 2024-10-23
APIGOV-29053 - support skipping handling external CRDs

dgghinea:
- adds an interface with which the agent can send what CRDs to ignore handling

### DIFF
--- a/pkg/apic/provisioning/definitions.go
+++ b/pkg/apic/provisioning/definitions.go
@@ -9,6 +9,7 @@ const (
 	OAuthSecretCRD    = "oauth-secret"
 	OAuthPublicKeyCRD = "oauth-public-key"
 	OAuthIDPCRD       = "oauth-idp"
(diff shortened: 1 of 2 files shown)


pkg/apic/provisioning/definitions.go
```diff
@@ -129,6 +130,10 @@ type Provisioning interface {
 	CredentialUpdate(CredentialRequest) (RequestStatus, Credential)
 }
 
+type CustomCredential interface {
+	GetIgnoredCredentials() []string
```
vivekschauhan, line 134: can we change this to GetIgnoredCredentialTypes ?

dgghinea: we are ignoring credentials by name, right? should this be CredentialNames?

vivekschauhan (approved):

jcollins-axway (approved):

PR #849, opened 2024-10-31
APIGOV-28984 [SDK] Custom unit configuration and client code - Initial Checkin

kasuvy:
I will keep adding some more tests meanwhile, please take a look and give me suggestions on improving this. Thanks in advance.


pkg/agent/provisioning.go
```diff
@@ -537,7 +537,7 @@ func RegisterProvisioner(provisioner provisioning.Provisioning) {
 	if agent.cfg.GetAgentType() == config.DiscoveryAgent || agent.cfg.GetAgentType() == config.GovernanceAgent {
 		agent.proxyResourceHandler.RegisterTargetHandler(
 			"accessrequesthandler",
-			handler.NewAccessRequestHandler(agent.provisioner, agent.cacheManager, agent.apicClient),
```
jcollins-axway, line 540: Maybe instead of sending the config in here, we can create a handler in the customunit package. Then that handler is sent in and all the logic in the access request handler will be processed in the new custom unit quota handler?

pkg/customunit/client.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+			c.MetricReporting()
+		case <-c.metricReportingServiceClient.Context().Done():
+			if c.isRunning {
+				c.isRunning = false
+				c.timer.Stop()
+				c.cancelCtx()
+			}
+			c.MetricReporting()
+		case <-c.timer.C:
```
jcollins-axway, line 176: Can this be changes to 

```go
case report, err := <-c.metricReportingServiceClient.Recv():
  if err == io.EOF {
    c.logger.Debug("stream finished")
		break
  }
  if err != nil {
    break
  }
  c.reportMetrics(metricReport)
```

pkg/customunit/client.go
```diff
@@ -0,0 +1,332 @@
...
+		apiSvc = c.cache.GetAPIServiceWithName(apiSvcValue)
+	}
+	if apiSvc == nil {
+		return nil, nil
+	}
+
+	return &models.APIDetails{
+		ID:     apiSvc.Metadata.ID,
+		Name:   apiSvc.Name,
+		TeamID: apiSvc.Owner.ID,
```
jcollins-axway, line 281: TeamID will not be used anymore in the new metric, so you can drop it

vivekschauhan (approved):

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+		cache:            cache,
+		agentType:        agentType,
+		metricReportChan: make(chan *customunits.MetricReport, 100),
+		stopChan:         make(chan struct{}),
+		clients:          []*customUnitClient{},
+		logger:           log.NewFieldLogger().WithPackage("customunit").WithComponent("manager"),
+	}
+}
+
+func (h *CustomUnitMetricServerManager) HandleQuotaEnforcement(ctx context.Context, cancelCtx context.CancelFunc, ar *management.AccessRequest, app *management.ManagedApplication) error {
```
jcollins-axway, line 48: does not look like ctx or cancelCtx are used in the func

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+	instRef := ar.GetReferenceByGVK(management.APIServiceInstanceGVK())
+	instID := instRef.ID
+	instance, err := h.cache.GetAPIServiceInstanceByID(instID)
+	if err != nil {
+		return nil, err
+	}
+	return instance, nil
+}
+
+func (m *CustomUnitMetricServerManager) HandleMetricReporting(ctx context.Context, cancelCtx context.CancelFunc, metricCollector metricCollector) {
```
jcollins-axway, line 152: does not look like ctx or cancelCtx are used in the func here either

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+	if m.agentType != config.TraceabilityAgent {
+		return
+	}
+	go m.receiveMetrics(metricCollector)
```
jcollins-axway, line 156: maybe only start this go routine is their are metric services we are connecting to

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+	for {
+		select {
+		case metricReport := <-c.metricReportChan:
+			if metricReport == nil {
+				continue
+			}
+			logger := c.logger.WithField("api", metricReport.ApiService.GetValue())
+			customMetricDetail, err := c.buildCustomMetricDetail(metricReport)
+			if err != nil {
+				logger.Error(err)
```
jcollins-axway, line 177: you can change this to logger.WithError(err).Error("could not build metric data")

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func (c *CustomUnitMetricServerManager) receiveMetrics(metricCollector metricCollector) {
+	for {
+		select {
+		case metricReport := <-c.metricReportChan:
+			if metricReport == nil {
+				continue
+			}
+			logger := c.logger.WithField("api", metricReport.ApiService.GetValue())
```
jcollins-axway, line 174: maybe other metric fields can be added here too, given they are in the metric

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+	management "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/management/v1alpha1"
+	"github.com/Axway/agent-sdk/pkg/apic/definitions"
+	"github.com/Axway/agent-sdk/pkg/config"
+	"github.com/Axway/agent-sdk/pkg/transaction/models"
+	transUtil "github.com/Axway/agent-sdk/pkg/transaction/util"
+	"github.com/Axway/agent-sdk/pkg/util"
+	"github.com/Axway/agent-sdk/pkg/util/log"
+)
+
+type CustomUnitMetricServerManager struct {
```
jcollins-axway, line 22: Maybe just call this CustomUnitHandler since it handles quota and metrics

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+	configs          []config.MetricServiceConfiguration
```
jcollins-axway, line 23: configs seems too generic, servicesConfigs?

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+	cache            cache.Manager
```
jcollins-axway, line 24: rather than the whole cache.Manager this should use a locally defined interface with just the methods in the cache that are used

pkg/customunit/manager.go
```diff
@@ -0,0 +1,331 @@
...
+func (h *CustomUnitMetricServerManager) HandleQuotaEnforcement(ctx context.Context, cancelCtx context.CancelFunc, ar *management.AccessRequest, app *management.ManagedApplication) error {
+	// Build quota info
+	quotaInfo, err := h.buildQuotaInfo(ctx, ar, app)
+	if err != nil {
+		return fmt.Errorf("could not build quota info from access request")
+	}
+	errMessage := ""
+	for _, config := range h.configs {
+		if config.MetricServiceEnabled() {
+			factory := NewCustomUnitClientFactory(config.URL, h.cache, quotaInfo)
```
jcollins-axway, line 57: We should have the factories created in the constructor with things that do not change on each connection, such as the cache that is sent in and any grpc opts that are used. Then the factory will receive only the args that differ between services or calls, url, and quotaInfo here.

Similar for the metric service factory below

pkg/customunit/client.go
```diff
@@ -0,0 +1,145 @@
...
+		c := &customUnitClient{
+			quotaInfo: quotaInfo,
+			url:       url,
+			dialOpts: []grpc.DialOption{
+				grpc.WithTransportCredentials(insecure.NewCredentials()),
+			},
+			cache:     agentCache,
+			logger:    log.NewFieldLogger().WithPackage("customunit").WithComponent("client").WithField("metricServer", url),
+			stopChan:  make(chan struct{}),
+			isRunning: true,
```
jcollins-axway, line 42: isRunning should probably be false on init?

pkg/customunit/client.go
```diff
@@ -0,0 +1,145 @@
...
+
+type customUnitClient struct {
+	logger    log.FieldLogger
+	quotaInfo *cu.QuotaInfo
+	dialOpts  []grpc.DialOption
+	cOpts     []grpc.CallOption
+	url       string
+	conn      *grpc.ClientConn
+	isRunning bool
+	cache     cache.Manager
```
jcollins-axway, line 22: same as my other comment, lets use a local interface with just the funcs we need from the cache manager

pkg/customunit/client.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func WithGRPCDialOption(opt grpc.DialOption) CustomUnitOption {
+	return func(c *customUnitClient) {
+		c.dialOpts = append(c.dialOpts, opt)
+	}
+}
+
+func (c *customUnitClient) createConnection() error {
+	conn, err := grpc.DialContext(context.Background(), c.url, c.dialOpts...)
```
jcollins-axway, line 61: Maybe we do not need the grpc.DialContext call since we are not using the context for these

pkg/customunit/client.go
```diff
@@ -0,0 +1,145 @@
...
+	return client.QuotaEnforcementInfo(context.Background(), c.quotaInfo, c.cOpts...)
+}
+
+func (c *customUnitClient) StartMetricReporting(metricReportChan chan *cu.MetricReport) {
+	c.isRunning = false
+	for {
+		err := c.createConnection()
+		if err != nil {
+			c.ExecuteBackoff()
+			if c.delay > 5*time.Minute {
```
jcollins-axway, line 84: I do not think we want to break in this case, this should just be the max delay. Lets move this up as a constant at the top of the file.

pkg/customunit/client.go
```diff
@@ -0,0 +1,145 @@
...
+			metricReportChan <- metricReport
+		}
+	}
+
+}
+
+func (c *customUnitClient) ExecuteBackoff() {
+	c.logger.Debugf("connection is still lost, trying again in %v.", c.delay)
+	time.Sleep(c.delay)
+	c.delay = 2 * c.delay
```
jcollins-axway, line 134: Check is the max backoff delay was hit here and set to it

pkg/customunit/client.go
vivekschauhan: DialContext is deprecated and replaced with grpc.NewClient()

jcollins-axway (approved):

vivekschauhan (approved):

jcollins-axway (approved):

PR #864, opened 2024-12-04
APIGOV-27681 - event generator updates for processing

jcollins-axway:
- create event report interface and builder to help create it
- add new processor to receive the event report and create transaction events
- use event report in old create event methods to synchronize logic of all processors
- cleanup from linting warnings


pkg/transaction/eventgenerator.go
```diff
@@ -59,19 +61,127 @@ func (e *Generator) SetUseTrafficForAggregation(useTrafficForAggregation bool) {
...
+
+	events, err := e.CreateFromEventReport(report)
+	if err != nil {
+		return beat.Event{}, err
+	}
+	if len(events) == 0 {
+		return beat.Event{}, errors.New("an event was not created")
+	}
+	if len(events) > 1 {
+		return events[0], errors.New("unexpectedly, more than one event was created, only returning the first")
```
sbolosan, line 91: since we are getting only the first one, regardless of if theres more than 1, should this just warn in the if statement, and then fall through to the final return?

sbolosan: ...unless you really really wanted to return an error?

sbolosan (approved):

pkg/transaction/eventgenerator.go
jcollins-axway: that would be the agent implementations issue here

vivekschauhan (approved):

PR #867, opened 2024-12-09
APIGOV-29284 - agent subresource hashes

dgghinea:
- moved func(currently not used anymore) from definitions to util pkg
- hashes are stored in a field from ResourceMeta(outside of api-server) and inside x-agent-details on the api-server resource
- createHashes is called after getting a cached resource for every resource type


jcollins-axway:
It looks like the SDK has to explicitly call to create these hashes, was there a reason that this path was taken? Not sure if I just don't remember this discussion.

I also thought we discussed putting this in the x-agent-details as well, has that implementation changed? If this is what we have to do then all gucci.

jcollins-axway (approved):

dgghinea:
> It looks like the SDK has to explicitly call to create these hashes, was there a reason that this path was taken? Not sure if I just don't remember this discussion.
- maybe I'm not totally getting it, but who else should call to create the hashes? 


> I also thought we discussed putting this in the x-agent-details as well, has that implementation changed? If this is what we have to do then all gucci.
- probably I should have left some comments regarding this, but the path that I chose was: keep the hashes in a field(ignored for marshal-unmarshal) from ResourceMeta, but before the SubResource call is made, move them inside x-agent-details. After the call is finished, delete from x-agent-details and move back to that field. I made it like this because there are multiple places where the agent details are overwritten and the hashes would have been lost.

alrosca (approved):

PR #876, opened 2025-01-21
APIGOV-29284-2 - fix some hash misses + codegen

dgghinea:
Copies the SubResourceHashes when using ri.FromInstance() and ri.AsInstance() because some resources, when gathered the cache, need to be converted


pkg/apic/apiserver/models/management/v1alpha1/APIService.go
```diff
@@ -110,7 +110,7 @@ func (res *APIService) AsInstance() (*apiv1.ResourceInstance, error) {
 	if err != nil {
```
jcollins-axway, line 110: everything in this folder path is generated code, the code within should be added to the template and not directly here

PR #885, opened 2025-02-03
APIGOV-29684 - Update startup logic order of operations

jcollins-axway:
- Update the startup logic in regards to IDP, Cred Mapping, and Marketplace Resource Handling

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -73,6 +73,7 @@ type agentData struct {
 	agentShutdownHandler       ShutdownHandler
 	proxyResourceHandler       *handler.StreamWatchProxyHandler
 	isInitialized              bool
+	isFinalized                bool
 
 	provisioner          provisioning.Provisioning
 	streamer             *stream.StreamerClient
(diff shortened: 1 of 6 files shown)


pkg/agent/agent.go
```diff
@@ -239,10 +240,25 @@ func InitializeProfiling(cpuProfile, memProfile string) {
...
+
+func finalizeInitialization() error {
+	if agent.isInitialized {
+		return nil
+	}
+
 	err := registerExternalIDPs()
 	if err != nil {
-		logger.WithError(err).Error("failed to register CRDs for external IdP config")
+		logger.WithError(err).Fatal("failed to register CRDs for external IdP config")
```
jcollins-axway, line 254: This registration only happen on init, if we can not register all the IDPs then the agent should stop. This will avoid cases where a failed IDP registration causes a CRD to be modified.

vivekschauhan (approved):

vivekschauhan (approved):

sbolosan (approved):

PR #887, opened 2025-02-10
APIGOV-29633 - updates for application profile provisioning

jcollins-axway:
- Add new builder method for adding enum labels and values
- Update provisioning to map selected label in request to mapped value
- Create mock type for app profile def testing

### DIFF
--- a/pkg/agent/handler/managedapplicationprofile.go
+++ b/pkg/agent/handler/managedapplicationprofile.go
@@ -147,7 +147,6 @@ func (h *managedApplicationProfile) getManagedApp(_ context.Context, profile *ma
 		return nil, err
 	}
 
-	app = &management.ManagedApplication{}
 	err = app.FromInstance(ri)
 	return app, err
 }
--- a/pkg/agent/handler/managedapplicationprofile.go
+++ b/pkg/agent/handler/managedapplicationprofile.go
@@ -119,7 +119,7 @@ func (h *managedApplicationProfile) onPending(ctx context.Context, profile *mana
 }
 
 func (h *managedApplicationProfile) checkForEnumValueMap(_ context.Context, data map[string]interface{}, profileDef string) {
-	log := h.logger.WithField("applicaitonProfileDefinition", profileDef)
+	log := h.logger.WithField("applicationProfileDefinition", profileDef)
 
 	// get application profile definition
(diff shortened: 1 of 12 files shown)


pkg/agent/handler/managedapplicationprofile.go
```diff
@@ -117,6 +118,40 @@ func (h *managedApplicationProfile) onPending(ctx context.Context, profile *mana
...
+	log := h.logger.WithField("applicationProfileDefinition", profileDef)
+
+	// get application profile definition
+	ri, err := h.cache.GetApplicationProfileDefinitionByName(profileDef)
+	if err != nil {
+		log.WithError(err).Error("error getting application profile definition from cache")
+		return
+	}
+	if ri == nil {
+		log.Debug("could not fine application profile definition in cache")
```
sbolosan, line 131: could not find

sbolosan (approved):

sbolosan (approved):

pkg/agent/handler/managedapplicationprofile.go
```diff
@@ -117,6 +118,40 @@ func (h *managedApplicationProfile) onPending(ctx context.Context, profile *mana
...
+		log.WithError(err).Error("error reading application profile definition from cache")
+		return
+	}
+
+	updateDataFromEnumMap(data, appProfDef.Spec.Schema)
+}
+
+func (h *managedApplicationProfile) getManagedApp(_ context.Context, profile *management.ManagedApplicationProfile) (*management.ManagedApplication, error) {
+	app := management.NewManagedApplication(profile.Spec.ManagedApplication, profile.Metadata.Scope.Name)
+	ri, err := h.client.GetResource(app.GetSelfLink())
```
vivekschauhan, line 145: Do we have to get the app from server or can we get it from cache?

jcollins-axway: I will give that a try, I had that there but hit an issue, will give it another go. Will change for creds and acc reqs too.

jcollins-axway: DAs do not cache managed applications. We should consider adding that, but in another story?

vivekschauhan (approved):

PR #896, opened 2025-02-26
APIGOV-29555 Timebox handling

alrosca:


pkg/agent/handler/agentresource.go
```diff
@@ -38,5 +46,16 @@ func (h *agentResourceHandler) Handle(ctx context.Context, _ *proto.EventMeta, r
 			h.agentResourceManager.SetAgentResource(resource)
 		}
 	}
+
+	if action == proto.Event_SUBRESOURCEUPDATED && resource.Kind == traceabilityAgent {
```
jcollins-axway, line 50: maybe we can also check that the agentstate subresource is also the subresource that changed?

perhaps do the opposite check to, fall out early if its not a subresource action or not the agentstate sub resource or not the traceability agnet kind?

vivekschauhan (approved):

jcollins-axway (approved):

sbolosan (approved):

PR #899, opened 2025-03-04
APIGOV-29802 int property uses lower and upper limits

alrosca:


dgghinea:
maybe some coverage for these? some examples should be in pkg/cmd/cmd.go

vivekschauhan (dismissed):

sbolosan (dismissed):

jcollins-axway (approved):

dgghinea (approved):

PR #903, opened 2025-03-13
APIGOV-29907 traceability trigger 

alrosca:


pkg/agent/handler/agentresource.go
```diff
@@ -60,6 +64,10 @@ func (h *agentResourceHandler) Handle(ctx context.Context, meta *proto.EventMeta
 	}
 	if ta.Agentstate.Sampling.Enabled {
 		h.sampler.EnableSampling(ta.Agentstate.Sampling.Limit, time.Time(ta.Agentstate.Sampling.EndTime))
+
+		if traceabilityTriggerHandler, ok := h.agentResourceManager.GetHandler().(TraceabilityTriggerHandler); ok {
```
vivekschauhan, line 68: Possible nil panic here if handler is not set

pkg/traceability/sampling/sample.go
```diff
@@ -112,3 +112,7 @@ func (s *sample) FilterEvents(events []publisher.Event) []publisher.Event {
 
 	return sampledEvents
 }
+
+func (s *sample) GetSamplePercentage() float64 {
```
vivekschauhan, line 116: Is this getting used somewhere ?

alrosca: Until we clean the percentage implementation, azure uses this method when creating ExecuteTraceability request

pkg/agent/handler/agentresource.go
alrosca: If handler is nil, ok will be false and won't execute the if statement block

jcollins-axway (approved):

PR #922, opened 2025-04-09
APIGOV-30183 - sanitize version over apiserver limit of 30 chars

sbolosan:
1. for oas2, oas3, and asyncapi - GetVersion()
2. util to take in version and verify length.
3. APIServer restricts length to 30 chars.  Util to check length and if length > 30, compute hash
4. Stringify hashInt to satisfy APIServer validation
5. If version length <= 30, NOP

### DIFF
--- a/pkg/apic/specoas3processor.go
+++ b/pkg/apic/specoas3processor.go
@@ -32,7 +32,7 @@ func (p *oas3SpecProcessor) GetResourceType() string {
 
 // GetVersion -
 func (p *oas3SpecProcessor) GetVersion() string {
-	return p.spec.Info.Version
+	return util.SanitizeServiceVersion(p.spec.Info.Version)
 }
 
 // GetEndpoints -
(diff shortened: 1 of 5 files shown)


pkg/apic/specoas3processor.go
```diff
@@ -32,7 +32,7 @@ func (p *oas3SpecProcessor) GetResourceType() string {
 
 // GetVersion -
 func (p *oas3SpecProcessor) GetVersion() string {
-	return p.spec.Info.Version
+	return util.SanitizeServiceVersion(p.spec.Info.Version)
```
vivekschauhan, line 35: Instead of sanitizing in parser, I would suggest to apply this in servicebodybuilder after the if block here(https://github.com/Axway/agent-sdk/blob/main/pkg/apic/servicebuilder.go#L312) to allow overridden versions as well to be sanitized

dgghinea (approved):

jcollins-axway (approved):

PR #926, opened 2025-04-24
APIGOV-29787 Update metrics reporting

alrosca:
Changed the way metrics are added in metrics collector. The event report and event generator are used now to add the metrics and the agent doesn't call the metrics collector directly, how it used to do.


pkg/transaction/eventgenerator.go
```diff
@@ -184,6 +147,36 @@ func (e *Generator) CreateFromEventReport(eventReport EventReport) ([]beat.Event
 	return append(events, detailEvents...), nil
 }
 
+func (e *Generator) AddMetricDetailsFromEventReport(eventReport EventReport) error {
```
jcollins-axway, line 150: Unfortunately this is not exactly what we were aiming for. The goal of the story was to receive many reported transactions in the event report then report them to the metrics collector so they can be aggregated together. In this implementation each metric will still be a separate event and not aggregated. 

- Send in the info about a batch of transactions to the event report, not necessarily using an existing metric structure
- Report into the metrics collector about that batch, either 
  - Breaking it down to call AddMetric XX transactions 
  - Receive the batch of transactions and let the collector handle aggregation

jcollins-axway (approved):

PR #929, opened 2025-04-30
APIGOV-30351 - updates for handling subresource spec hashing

jcollins-axway:
- Store as strings
- Remarshal to ensure consistent key ordering
- Check status err on apiservice to prevent updating time 
- Add test to cover remarshal func

### DIFF
--- a/pkg/apic/apiserver/models/api/v1/resourcemeta_test.go
+++ b/pkg/apic/apiserver/models/api/v1/resourcemeta_test.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 
 	"github.com/Axway/agent-sdk/pkg/util"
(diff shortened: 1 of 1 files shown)


sbolosan:
Do we have to think about any migration paths for spec hashes?

jcollins-axway:
> Do we have to think about any migration paths for spec hashes?

Good question, but no. Actually we probably need to dive a bit deeper. The agents actually do not use the subres hash that we store. They recalculate it just in time when retrieving from the cache.

vivekschauhan (approved):

pkg/apic/apiserver/models/api/v1/resourcemeta.go
```diff
@@ -179,17 +180,14 @@ func (rm *ResourceMeta) GetSubResource(key string) interface{} {
 	return rm.SubResources[key]
 }
 
-func (rm *ResourceMeta) GetSubResourceHash(key string) (float64, bool) {
+func (rm *ResourceMeta) GetSubResourceHash(key string) (string, bool) {
```
dgghinea, line 183: I think I've tried this approach with strings, but they failed due to the precision yielding out various results

PR #942, opened 2025-05-22
APIGOV-30507 - let the SDK marshall the spec from the processor, compact if necessary

jcollins-axway:


jcollins-axway:
Still need to figure out how to update tests with a large spec to hit the json.Compact calls.

jcollins-axway:
> Still need to figure out how to update tests with a large spec to hit the json.Compact calls.

with the testing I did json.Compact did not do anything different than the Marshal we were already doing, it is for compacting json data without the parsing it. Since we already parse the oas specs we can simply use marshal

dgghinea (approved):

vivekschauhan (approved):

alrosca (approved):

PR #945, opened 2025-06-04
APIGOV-30406 endpoints sampling support

alrosca:


pkg/agent/handler/agentresource.go
```diff
@@ -2,18 +2,23 @@ package handler
 
 import (
 	"context"
+	"strings"
 	"time"
 
+	agentcache "github.com/Axway/agent-sdk/pkg/agent/cache"
```
jcollins-axway, line 8: rather than importing the cache package, can you create a local interface with just the methods you need from the manager?

pkg/agent/handler/agentresource.go
```diff
@@ -118,10 +132,59 @@ func (h *agentResourceHandler) handleCompliance(action proto.Event_Type, subres
...
+			h.logger.Warnf("API Service Instance %s does not have agent details", apiSIRI.Metadata.ID)
+			continue
+		}
+		apiIDI, ok := apiSIAgentDetails[definitions.AttrExternalAPIID]
+		if !ok || apiIDI == "" {
+			h.logger.Warnf("API Service Instance %s does not have external API ID", apiSIRI.Metadata.ID)
+			continue
+		}
+
+		apiID := apiIDI.(string)
```
dgghinea, line 171: I think this could be replaced with 
`apiID, err := skdUtil.GetAgentDetailsValue(apiSI, definitions.AttrExternalAPIID)`

pkg/agent/handler/agentresource.go
```diff
@@ -118,10 +132,59 @@ func (h *agentResourceHandler) handleCompliance(action proto.Event_Type, subres
...
 	}
 	return nil
 }
+
+func (h *agentResourceHandler) handleEndpointsSampling(endpoints []management.TraceabilityAgentAgentstateSamplingEndpoints) map[string]management.TraceabilityAgentAgentstateSamplingEndpoints {
+	endpointsInfo := make(map[string]management.TraceabilityAgentAgentstateSamplingEndpoints) // apiID -> endpoints
+	apiSIInfo := make(map[string]string)                                                      // basepath -> apiID
+
+	apiSIRIs := h.cache.ListAPIServiceInstances()
+	if len(apiSIRIs) == 0 {
```
dgghinea, line 148: also I don't think this len check is necessary, since the end result would be the same without it

pkg/traceability/sampling/sample.go
```diff
@@ -50,24 +49,17 @@ func (s *sample) EnableSampling(samplingLimit int32, samplingEndTime time.Time)
...
-}
-
-func (s *sample) DisableSampling() {
-	s.samplingLock.Lock()
-	defer s.samplingLock.Unlock()
-	if s.enabled {
-		s.disableSamplingCh <- struct{}{}
+	if s.endpointsSampling.enabled {
+		go s.disableEndpointsSampling()
+	} else {
```
dgghinea, line 54: rather unnecessary else due to the function ending here

dgghinea (approved):

pkg/agent/handler/agentresource.go
```diff
@@ -99,7 +111,9 @@ func (h *agentResourceHandler) handleTraceabilitySampling(resource *v1.ResourceI
 		return nil
 	}
 
-	h.sampler.EnableSampling(ta.Agentstate.Sampling.Limit, time.Time(ta.Agentstate.Sampling.EndTime))
+	endpointsInfo := h.handleEndpointsSampling(ta.Agentstate.Sampling.Endpoints)
+	h.sampler.EnableSampling(ta.Agentstate.Sampling.Limit, time.Time(ta.Agentstate.Sampling.EndTime), endpointsInfo)
```
dgghinea, line 115: also I'm not sure if this would work correctly when an endpoint is removed by the controller(given that there are 2 or more) as well as when there are multiple endpoints inside the TA.agentstate.sampling.endpoints

jcollins-axway (dismissed):

jcollins-axway (dismissed):

PR #946, opened 2025-06-10
APIGOV-30447 - Set the credential type in CredentialRequestDefinition

sbolosan:
The goal of the story is to have all Discovery Agents define the credential type in the CRD so that this information can be passed along to the Marketplace and help consumer understand their credentials

In case we cannot determine the credential type, leave the field empty.

### DIFF
--- a/scripts/apiserver/model_credential_request_definition_spec.tmpl
+++ b/scripts/apiserver/model_credential_request_definition_spec.tmpl
@@ -4,10 +4,11 @@ package management
 
 // CredentialRequestDefinitionSpec struct for CredentialRequestDefinitionSpec
 type CredentialRequestDefinitionSpec struct {
-	// JSON Schema draft \\#7 for defining the AccessRequest properties needed to get access to an APIServiceInstance.
-	Schema       map[string]interface{}                       `json:"schema"`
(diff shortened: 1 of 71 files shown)


scripts/apiserver/model_credential_request_definition_spec.tmpl
```diff
@@ -4,10 +4,11 @@ package management
 
 // CredentialRequestDefinitionSpec struct for CredentialRequestDefinitionSpec
 type CredentialRequestDefinitionSpec struct {
-	// JSON Schema draft \\#7 for defining the AccessRequest properties needed to get access to an APIServiceInstance.
-	Schema       map[string]interface{}                       `json:"schema"`
-	Provision    *CredentialRequestDefinitionSpecProvision    `json:"provision,omitempty"`
-	Webhooks     []CredentialRequestDefinitionSpecWebhook     `json:"webhooks,omitempty"`
+	// JSON Schema draft \\#7 for describing the fields needed to provision ManagedApplicationProfiles.
```
vivekschauhan, line 7: I suppose the comment was for Schema property

vivekschauhan (approved):

jcollins-axway (approved):

alrosca (approved):

PR #958, opened 2025-07-25
Apigov 30832

sbolosan:


pkg/transaction/metric/util.go
```diff
@@ -10,6 +10,11 @@ import (
 	transutil "github.com/Axway/agent-sdk/pkg/transaction/util"
 )
 
+const (
+	// UnknownAPIID - Constant for transactions with no API ID or API Name.  Exported to avoid cyclic issues with transaction package.
+	UnknownAPIID = "unknown-api-id"
```
dgghinea, line 15: now it was exported but the one in pkg/transaction/definitions.go still exists. I was hoping we could use a single const overall for "unknown-api-id"

pkg/transaction/metric/util.go
```diff
@@ -93,7 +98,7 @@ func centralMetricFromAPIMetric(in *APIMetric) *centralMetric {
 		}
 	}
 
-	if in.API.ID != unknown && in.API.ID != "" {
+	if in.API.ID != unknown && in.API.ID != UnknownAPIID && in.API.ID != "" {
```
jcollins-axway, line 101: Is unknown also defined in multiple places?

jcollins-axway:
Another question, do we think we need "unknown-api-id" or can we just use "unknown" what does the "api-id" part give us?

pkg/transaction/metric/util.go
sbolosan: Changed to using the existing 'unknown'... with this update, the unknown-api-id is not needed anymore.

jcollins-axway (approved):

PR #968, opened 2025-09-02
APIGOV-31098 - Add runtimeID to user-agent and update process to rebuild cache

sbolosan:
1. create runtime ID at startup
2. add it to the user agent
3. update process to rebuild cache
4. hardening to ensure cache is built properly

### DIFF
--- a/pkg/agent/resource/manager.go
+++ b/pkg/agent/resource/manager.go
@@ -163,10 +163,12 @@ func (a *agentResourceManager) UpdateAgentStatus(status, prevStatus, message str
 	// using discovery agent status here, but all agent status resources have the same structure
 	agentStatus := newDAStatus(agentInstance.ResourceMeta, status, prevStatus, message)
 
-	// See if we need to rebuildCache
-	timeToRebuild, _ := a.shouldRebuildCache()
-	if timeToRebuild && a.rebuildCache != nil {
-		a.rebuildCache.RebuildCache()
+	// Only perform periodic/self-healing cache rebuild for non-gRPC agents
+	if a.cfg != nil && !a.cfg.IsUsingGRPC() {
+		timeToRebuild, _ := a.shouldRebuildCache()
+		if timeToRebuild && a.rebuildCache != nil {
+			a.rebuildCache.RebuildCache()
+		}
 	}
 
 	subResources := make(map[string]interface{})
--- a/pkg/agent/resource/manager.go
+++ b/pkg/agent/resource/manager.go
@@ -163,12 +163,10 @@ func (a *agentResourceManager) UpdateAgentStatus(status, prevStatus, message str
 	// using discovery agent status here, but all agent status resources have the same structure
 	agentStatus := newDAStatus(agentInstance.ResourceMeta, status, prevStatus, message)
 
-	// Only perform periodic/self-healing cache rebuild for non-gRPC agents
-	if a.cfg != nil && !a.cfg.IsUsingGRPC() {
-		timeToRebuild, _ := a.shouldRebuildCache()
-		if timeToRebuild && a.rebuildCache != nil {
-			a.rebuildCache.RebuildCache()
-		}
+	// See if we need to rebuildCache
+	timeToRebuild, _ := a.shouldRebuildCache()
+	if timeToRebuild && a.rebuildCache != nil {
+		a.rebuildCache.RebuildCache()
 	}
 
 	subResources := make(map[string]interface{})
@@ -186,6 +184,11 @@ func (a *agentResourceManager) UpdateAgentStatus(status, prevStatus, message str
 // 1. On UpdateAgentStatus, if x-agent-details, key "cacheUpdateTime" doesn't exist or empty, rebuild cache to populate cacheUpdateTime
 // 2. On UpdateAgentStatus, if x-agent-details exists, check to see if its past 7 days since rebuildCache was ran.  If its pass 7 days, rebuildCache
 func (a *agentResourceManager) shouldRebuildCache() (bool, error) {
+	// Only perform periodic/self-healing cache rebuild for non-gRPC agents
+	if a.cfg != nil && a.cfg.IsUsingGRPC() {
+		return false, nil
+	}
+
 	rebuildCache := false
 	agentInstance := a.GetAgentResource()
 	agentDetails := agentInstance.GetSubResource(definitions.XAgentDetails)
--- a/pkg/util/useragent.go
+++ b/pkg/util/useragent.go
@@ -9,7 +9,7 @@ import (
 
 var (
 	agentInfoRe   = regexp.MustCompile(`^([a-zA-Z0-9]*)\/(\d*.\d*.\d*)[-]?([a-z0-9]*?) SDK/(\d*.\d*.\d.*) ([a-z][a-zA-Z0-9-]*) ([a-z][a-zA-Z0-9-.]*) (binary|docker)[ ]?(reactive)?`)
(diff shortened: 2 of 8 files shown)


pkg/util/useragent.go
```diff
@@ -22,16 +22,18 @@ type CentralUserAgent struct {
 	IsGRPC              bool   `json:"reactive"`
 	HostName            string `json:"hostname,omitempty"`
 	UseGRPCStatusUpdate bool   `json:"-"`
+	RuntimeID           string `json:"runtimeId,omitempty"`
```
jcollins-axway, line 25: the regexps at the top need some updating

pkg/agent/resource/manager.go
```diff
@@ -163,10 +163,12 @@ func (a *agentResourceManager) UpdateAgentStatus(status, prevStatus, message str
 	// using discovery agent status here, but all agent status resources have the same structure
 	agentStatus := newDAStatus(agentInstance.ResourceMeta, status, prevStatus, message)
 
-	// See if we need to rebuildCache
-	timeToRebuild, _ := a.shouldRebuildCache()
-	if timeToRebuild && a.rebuildCache != nil {
-		a.rebuildCache.RebuildCache()
+	// Only perform periodic/self-healing cache rebuild for non-gRPC agents
+	if a.cfg != nil && !a.cfg.IsUsingGRPC() {
+		timeToRebuild, _ := a.shouldRebuildCache()
```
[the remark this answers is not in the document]
sbolosan: you're right... in my case, if they are using gRPC, shouldREbuildCache is never called.

pkg/util/useragent.go
sbolosan: during my added tests, it seems that there are no current tests for the option of grpc (as part of the regex) - ??(grpc-go.*\/\d*.\d*.\d*)?$`).  I don't believe ?? is actual valid, no?

pkg/util/useragent.go
```diff
@@ -9,7 +9,7 @@ import (
 
 var (
 	agentInfoRe   = regexp.MustCompile(`^([a-zA-Z0-9]*)\/(\d*.\d*.\d*)[-]?([a-z0-9]*?) SDK/(\d*.\d*.\d.*) ([a-z][a-zA-Z0-9-]*) ([a-z][a-zA-Z0-9-.]*) (binary|docker)[ ]?(reactive)?`)
-	agentInfoReV2 = regexp.MustCompile(`^([a-zA-Z0-9]*)\/(\d*.\d*.\d*)[-]?([-a-z0-9A-Z]*?) \(sdkVer:(\d*.\d*.\d.*)\; env:([a-zA-Z0-9-]*)\; agent:([a-zA-Z0-9-.]*)\; reactive:(true|false)\; hostname:([a-zA-Z0-9-_.]*)\) ??(grpc-go.*\/\d*.\d*.\d*)?$`)
+	agentInfoReV2 = regexp.MustCompile(`^([a-zA-Z0-9]*)\/(\d*.\d*.\d*)[-]?([-a-z0-9A-Z]*?) \(sdkVer:(\d*.\d*.\d.*)\; env:([a-zA-Z0-9-]*)\; agent:([a-zA-Z0-9-.]*)\; reactive:(true|false)\; hostname:([a-zA-Z0-9-_.]*); runtimeID:([a-zA-Z0-9-_.]*)\) ??(grpc-go.*\/\d*.\d*.\d*)?$`)
```
jcollins-axway, line 12: This regexp update will work if all agents are updated right away, but if any agent is being used with this format and without a runtimeID it will not match.

I think something like this will work `^([a-zA-Z0-9]*)\/(\d*.\d*.\d*)[-]?([-a-z0-9A-Z]*?) \(sdkVer:(\d*.\d*.\d.*)\; env:([a-zA-Z0-9-]*)\; agent:([a-zA-Z0-9-.]*)\; reactive:(true|false)\; hostname:([a-zA-Z0-9-_.]*)(; runtimeID:([a-zA-Z0-9-_.]*))?\) ??(grpc-go.*\/\d*.\d*.\d*)?$`

https://regex101.com/r/BfVDD6/1

pkg/util/useragent_test.go
```diff
@@ -33,7 +34,7 @@ func TestFormatUserAgents(t *testing.T) {
 			agentName:         "agent",
 			agentVersion:      "v1.0.0-125678e",
 			sdkVersion:        "v1.1.100",
-			expectedUserAgent: fmt.Sprintf("Test/1.0.0-125678e (sdkVer:1.1.100; env:env; agent:agent; reactive:true; hostname:%s)", hostname),
+			expectedUserAgent: fmt.Sprintf("Test/1.0.0-125678e (sdkVer:1.1.100; env:env; agent:agent; reactive:true; hostname:%s; runtimeID:%s)", hostname, ""),
```
jcollins-axway, line 37: I do not think a test with an empty runtimeID is necessary.

However some tests with the runtimeID part of the user agent and without would be beneficial.

pkg/util/useragent_test.go
```diff
@@ -187,11 +178,12 @@ func TestParseUserAgents(t *testing.T) {
 				IsGRPC:              true,
 				HostName:            hostname,
 				UseGRPCStatusUpdate: true,
+				RuntimeID:           "",
```
jcollins-axway, line 181: same as above, no reason to have a test with an empty runtimeID, but parsing tests without the section make sense

pkg/util/useragent.go
sbolosan: gotchu... i will resolve the other comment.  but the question still stands.  What about the ??(grpc-go.....) .  Is this valid?  I don't think ?? is a valid optional regex syntax

pkg/util/useragent.go
sbolosan: resolving this for a better regex provided by jason

jcollins-axway (approved):

pkg/util/useragent.go
```diff
@@ -9,7 +9,7 @@ import (
 
 var (
 	agentInfoRe   = regexp.MustCompile(`^([a-zA-Z0-9]*)\/(\d*.\d*.\d*)[-]?([a-z0-9]*?) SDK/(\d*.\d*.\d.*) ([a-z][a-zA-Z0-9-]*) ([a-z][a-zA-Z0-9-.]*) (binary|docker)[ ]?(reactive)?`)
-	agentInfoReV2 = regexp.MustCompile(`^([a-zA-Z0-9]*)\/(\d*.\d*.\d*)[-]?([-a-z0-9A-Z]*?) \(sdkVer:(\d*.\d*.\d.*)\; env:([a-zA-Z0-9-]*)\; agent:([a-zA-Z0-9-.]*)\; reactive:(true|false)\; hostname:([a-zA-Z0-9-_.]*)\) ??(grpc-go.*\/\d*.\d*.\d*)?$`)
+	agentInfoReV2 = regexp.MustCompile(`^([a-zA-Z0-9]*)\/(\d*.\d*.\d*)[-]?([-a-z0-9A-Z]*?) \(sdkVer:(\d*.\d*.\d.*)\; env:([a-zA-Z0-9-]*)\; agent:([a-zA-Z0-9-.]*)\; reactive:(true|false)\; hostname:([a-zA-Z0-9-_.]*)(; runtimeID:([a-zA-Z0-9-_.]*))?\) ??(grpc-go.*\/\d*.\d*.\d*)?$`)
```
vivekschauhan, line 12: can we change runtimeID to runtimeId?

pkg/util/useragent.go
```diff
@@ -22,29 +22,39 @@ type CentralUserAgent struct {
...
-			reactive = "true"
-		}
-		ua = fmt.Sprintf("%s/%s (sdkVer:%s; env:%s; agent:%s; reactive:%s; hostname:%s)", ca.AgentType, ca.Version, ca.SDKVersion, ca.Environment, ca.AgentName, reactive, hostName)
+	reactive := "false"
+	if ca.IsGRPC {
+		reactive = "true"
 	}
+
+	ua := fmt.Sprintf(
+		"%s/%s (sdkVer:%s; env:%s; agent:%s; reactive:%s; hostname:%s; runtimeID:%s)",
```
vivekschauhan, line 48: runtimeID to runtimeId, here as well

vivekschauhan (approved):

PR #971, opened 2025-09-16
APIGOV-31178 - pass context from manager to clients for cancelling

jcollins-axway:
- Use an atomic value in place of a mutex
- Update stream and poll client managers to create context/cnacelFunc and send to its clients
- Some clients can cancel the context, triggering all clients to stop
- All clients should check the context prior to writing to any channels, ending any loops if a cancellation is determined

### DIFF
--- a/pkg/agent/events/requestqueue.go
+++ b/pkg/agent/events/requestqueue.go
@@ -74,8 +74,6 @@ func (q *requestQueue) Write(request *proto.Request) error {
 
 func (q *requestQueue) Start() {
 	go func() {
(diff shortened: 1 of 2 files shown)


sbolosan (dismissed):
solid... better cancellation semantics and hopefully plugging any race conditions.
looks good!

sbolosan (dismissed):

dgghinea (approved):

sbolosan (approved):

vivekschauhan (approved):

PR #972, opened 2025-09-18
APIGOV-31195 - usage publisher to wait for ongoing metric collection before processing

jcollins-axway:


pkg/transaction/metric/metricscollector.go
```diff
@@ -210,6 +210,12 @@ func (c *collector) Ready() bool {
 	return agent.GetCentralConfig().GetEnvironmentID() != ""
 }
 
+// metricCheck - lock to wait for any inflight metric processing to complete
+func (c *collector) metricCheck() {
```
dgghinea, line 214: pretty smart way to do this 😄

dgghinea (approved):

PR #1015, opened 2026-03-10
APIGOV-31191 validate cache

alrosca:
Cache validation on start up and reconnect


jcollins-axway (dismissed):
Looks good from a high level, needs pounded testing wise

pkg/agent/handler/discoveryaccessrequest_test.go
```diff
@@ -0,0 +1,132 @@
...
+		Status: &apiv1.ResourceStatus{
+			Level: "Success",
+		},
+	}
+
+	ri, _ := ar.AsInstance()
+	// deleting state with success status - should NOT be cached
+	err := handler.Handle(NewEventContext(proto.Event_CREATED, nil, ri.Kind, ri.Name), nil, ri)
+	assert.Nil(t, err)
+	assert.Equal(t, []string{}, cm.GetAccessRequestCacheKeys())
```
jcollins-axway: as we spoke of we prefer test cases over single tests

jcollins-axway (dismissed):

jcollins-axway (dismissed):

pkg/agent/resource/manager.go
sbolosan: should we eat the error? 
At least log the error/warn and then if requirement is to rebuild, then rebuild?

sbolosan: I know this is old and not part of your changes, but can we do a safe guard here
```
strVal, ok := value.(string)
            if !ok {
                logger.Warn("cacheUpdateTime is not a string, triggering rebuild")
                return true, nil
            }

**Review by jasonmscollins — CHANGES_REQUESTED** [review]
See comments from Shane

**Review by jcollins-axway — CHANGES_REQUESTED** [review]
see comments

File: pkg/agent/stream/client.go [hunk 2]
```diff
@@ -172,6 +175,10 @@ func (s *StreamerClient) Start() error {
 	if s.onStreamConnection != nil {
 		s.onStreamConnection()
 	}
+	if s.onReconnect != nil && !s.firstStart {
```
jcollins-axway: rather than using a firstStart boolean, could we use sync.Once?

pkg/agent/cachevalidationjob.go
```diff
@@ -0,0 +1,172 @@
...
+		kinds[management.ManagedApplicationGVK().Kind] = struct{}{}
+		kinds[management.AccessRequestGVK().Kind] = struct{}{}
+		kinds[management.AccessRequestDefinitionGVK().Kind] = struct{}{}
+		kinds[management.CredentialRequestDefinitionGVK().Kind] = struct{}{}
+		kinds[management.ApplicationProfileDefinitionGVK().Kind] = struct{}{}
+	case config.TraceabilityAgent:
+		kinds[management.ManagedApplicationGVK().Kind] = struct{}{}
+		kinds[management.AccessRequestGVK().Kind] = struct{}{}
+	case config.ComplianceAgent:
+		kinds[management.ComplianceRuntimeResultGVK().Kind] = struct{}{}
```
jcollins-axway: any way to use the watch topic that we create to determine these types?

alrosca: the watch topic includes kinds that are not persisted in cache (subscribed for event processing only), so deriving the validated set purely from the watch topic would cause false-positive validation failures for those kinds. The per-agent-type whitelist is intentional.

pkg/agent/stream/client.go
alrosca: sync.Once doesn't fit here because the semantics are inverted — sync.Once runs a function exactly once, but the intent is the opposite: skip on first call, run on all subsequent calls. Expressing that with sync.Once requires an awkward closure that is harder to read than the bool.

Thread safety is also already covered: Start() guards entry with isRunning (atomic.Bool), so concurrent calls are rejected before firstStart is ever touched. There's no race to protect against.

pkg/agent/cachevalidationjob.go
jcollins-axway: The agent creates the watch topic on start, the logic that creates that has all the kinds, that is what I refer to. Not necessarily pulling the watch topic itself.

pkg/agent/resource/manager.go
sbolosan: Technically you're right. But one extra ok check costs nothing and eliminates an entire class of runtime panic, no? I'm just saying x-agent-details is a map deserialized from JSON and JSON deseralization into interface{} has no type gurantees. It'll panic if future code changes forgets to stringify it.  Up to you.

pkg/agent/eventsync.go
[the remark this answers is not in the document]
jcollins-axway: lets put something on the backlog to take a look @sbolosan

jcollins-axway (dismissed):

jcollins-axway (approved):

PR #1020, opened 2026-03-19
APIGOV-31809 metrics min, max, avg preserving

alrosca:


jcollins-axway (changes requested):
Could we add some unit tests around this and validate the min, max, average with various counts?

sbolosan:
> Could we add some unit tests around this and validate the min, max, average with various counts?

You could pull in the logging content from the parent ticket that actually has the real values to put through the unit tests

sbolosan (approved):

jcollins-axway (approved):

PR #1023, opened 2026-04-01
APIGOV-32372 cache rebuild optimizations

alrosca:
Commits:

- rebuild only failed kinds
- fix for listener/cache rebuild race condition
- fix to wait for clients (stream/poll) start and listen before healthcheck marks agent as ready


pkg/agent/stream/client.go
```diff
@@ -216,6 +216,23 @@ func (s *StreamerClient) Stop() {
...
+}
+
+// ResumeListener resumes event processing after a PauseListener call.
+// No-op when called before the listener has been created.
+func (s *StreamerClient) ResumeListener() {
+	if s.listener != nil {
+		s.listener.Resume()
+	}
+}
+
```
jcollins-axway, line 263: we will have to sync changes with stuff Vivek is working on now, so some stuff in here may change, especially in the connection to watch controller

pkg/agent/cachevalidationjob.go
```diff
@@ -45,17 +46,39 @@ func newCacheValidator(
 	}
 }
 
-func (cv *cacheValidator) Execute() error {
+// Execute validates each filter in the watch topic against the API server.
+// It returns the slice of filters whose cache is out of sync, plus errCacheOutOfSync
+// if any failed, or a nil slice and nil error if all filters are in sync.
+func (cv *cacheValidator) Execute() ([]management.WatchTopicSpecFilters, error) {
```
jcollins-axway, line 48: is the cacheValidator not a job? With this change does it not fail when registering it with the job pool?

pkg/agent/eventsync.go
```diff
@@ -131,11 +140,30 @@ func (es *EventSync) registerInstanceValidator() error {
...
+			agent.cacheManager.FlushKind(f.Kind)
+		}
+		if rebuildErr := es.discoveryCache.execute(failedFilters...); rebuildErr == nil {
+			if seqID > 0 {
+				es.sequence.SetSequence(seqID - 1)
+			}
+			agent.cacheManager.SaveCache()
+			es.resetCacheTimer()
+			return nil
+		} else {
```
jcollins-axway: I think we can drop this else, the if will return

jcollins-axway (changes requested):
see comments

pkg/agent/cachevalidationjob.go
alrosca: Is not a job that implements Ready() and Status() because is not bound by time, but by external events like start up or reconnect. It executes on demand, the only time when it runs bound by time is for cache rebuild check, but that is another job that triggers cache validation. I think this is the right approach since cache rebuild needs a synchronous response.

pkg/agent/stream/client.go
alrosca: These optimizations are not urgent and cache validation can function well enough without them, so I propose that Vivek's changes get merged before this and I'll adapt these optimizations around Vivek's changes

jcollins-axway (approved):
Changes look good to me, just need a lot of testing.

pkg/agent/eventsync.go
```diff
@@ -163,24 +195,36 @@ func (es *EventSync) resetCacheTimer() {
...
 	PublishingLock()
 	defer PublishingUnlock()
 
-	es.waitForCacheRebuild()
+	es.waitForCacheRebuild(filters...)
+}
+
+// pausedInitCache pauses the event listener, calls initCache, then resumes via defer.
+// A separate function ensures ResumeListener is always called even on panic.
+func (es *EventSync) pausedInitCache(filters ...management.WatchTopicSpecFilters) error {
```
sbolosan, line 216: found during testing. Worth while looking - fatal error: sync: Unlock of unlocked RWMutex
goroutine [running]:
sync.(*RWMutex).Unlock
github.com/Axway/agent-sdk/pkg/agent/events.(*EventListener).Resume (eventlistener.go:81)
github.com/Axway/agent-sdk/pkg/agent/stream.(*StreamerClient).ResumeListener (client.go:263)
github.com/Axway/agent-sdk/pkg/agent.(*EventSync).pausedInitCache (eventsync.go:219)
github.com/Axway/agent-sdk/pkg/agent.(*EventSync).RebuildCache (eventsync.go:208)
github.com/Axway/agent-sdk/pkg/agent.(*EventSync).validateAndRebuildCache (eventsync.go:271)

pkg/agent/stream/client.go
```diff
@@ -212,11 +226,44 @@ func (s *StreamerClient) Status() error {
...
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// PauseListener suspends event processing on the underlying listener.
+// Blocks until any in-flight event is done, then holds the pause until ResumeListener is called.
+// No-op when called before the listener has been created.
+func (s *StreamerClient) PauseListener() {
```
sbolosan: connected with panic issue?
When pausedInitCache is called on reconnect, looks like PauseListener is a no-op because s.listener is nil at that point. So no lock is acquired, right? The deferred ResumeListener then runs after the listener has been initialized, calling Unlock on a mutex that was never locked.

Just quick thoughts.  I'll let you figure it out.  But maybe we can have PauseListener return a bool indicating whether the pause was applied, and only call ResumeListener if it was. Just a quick thought.

dgghinea (approved):

jcollins-axway (approved):

PR #1026, opened 2026-04-16
APIGOV-32465 - Fix to avoid deadlocks on watch request and event channel

vivekschauhan:


pkg/agent/poller/poller.go
```diff
@@ -50,32 +50,28 @@ func newPollExecutor(interval time.Duration, options ...executorOpt) *pollExecut
 }
 
 // RegisterWatch registers a watch topic for polling events and publishing events on a channel
-func (m *pollExecutor) RegisterWatch(eventChan chan *proto.Event, errChan chan error) {
+func (m *pollExecutor) RegisterWatch(eventChan chan *proto.Event) {
```
jcollins-axway, line 53: love how this func was simplified

pkg/watchmanager/manager.go
```diff
@@ -135,13 +135,20 @@ func (m *watchManager) eventCatchUp(link string, events chan *proto.Event) error
...
 
 // RegisterWatch - Registers a subscription with watch service using topic
-func (m *watchManager) RegisterWatch(link string, events chan *proto.Event, errors chan error) (string, error) {
+func (m *watchManager) RegisterWatch(link string, events chan *proto.Event) (string, error) {
+	subscriptionID, _ := uuid.NewUUID()
+	subID := subscriptionID.String()
+
+	m.logger.
+		WithField("watchId", subID).
+		WithField("watchTopic", link).
```
jcollins-axway: Should we not just create a local logger here with the fields added then reuse throughout the func?

jcollins-axway (dismissed):

jcollins-axway (approved):

PR #1027, opened 2026-04-22
APIGOV-32445 409 resource conflict handling

alrosca:


pkg/apic/client.go
```diff
@@ -915,12 +915,32 @@ func (c *ServiceClient) updateSpecORCreateResourceInstance(data *apiv1.ResourceI
...
-		response, err := c.ExecuteAPI(method, url, nil, reqBytes)
+		response, err := c.executeAPI(method, url, nil, reqBytes, nil)
 		if err != nil {
 			return nil, err
 		}
 
-		err = json.Unmarshal(response, newRI)
+		// Handle 409 Conflict on POST: resource exists on server but not in cache.
+		// Fetch the existing resource, add it to the cache, and return it.
+		if response.Code == http.StatusConflict && method == coreapi.POST {
```
dgghinea: I would personally do a different approach. Leave L918 as it is, but add another case in the ExecuteAPIWithHeader() for 409 status code, returning (nil, ErrConflict)
Then, in this func, first do if err == ErrConflict and do the L926-L932 part
And simply remove L935-L943 that's already handled by the ExecuteAPI()

vivekschauhan (dismissed):

dgghinea (dismissed):

jcollins-axway (dismissed):

pkg/apic/client.go
```diff
@@ -915,12 +917,22 @@ func (c *ServiceClient) updateSpecORCreateResourceInstance(data *apiv1.ResourceI
...
+		respBytes, err := c.ExecuteAPIWithHeader(method, url, nil, reqBytes, nil)
+		if err == ErrConflict && method == coreapi.POST {
+			// Resource exists on server but not in cache; fetch and cache it.
+			log.Warnf("resource %s %s already exists on the server but was not in the local cache, fetching existing resource", data.Kind, data.Name)
+			fetchedRI, fetchErr := c.GetResource(data.GetSelfLink())
+			if fetchErr != nil {
+				return nil, fetchErr
+			}
+			c.addResourceToCache(fetchedRI)
+			return fetchedRI, nil
```
dgghinea: I just realised. Wouldn't this skip the agent-details update? I think we need to do 
`newRI = fetchedRI`
and continue processing

dgghinea (approved):

PR #1029, opened 2026-04-28
APIGOV-32516 - Create IdentityProvider and IdentityProviderMetadata resources from agent IdP config

sbolosan:
New feature flag agentFeatures.manageIDPResources (default: false). When enabled, the SDK creates and maintains IdentityProvider and IdentityProviderMetadata resources in Engage during IdP provider registration.

IdP lifecycle logic queries Engage before creating to avoid duplicates. On creation, environment credential policies (expiry/visibility) are applied to the new resource and IdentityProviderMetadata is populated from live IdP server metadata.

Optional IDPResourceSupplier extension point allows agent implementations to supply custom IdentityProvider or IdentityProviderMetadata resource shapes instead of the SDK defaults.

The CredentialRequestDefinition registered at startup is linked to the IdentityProvider resource name created or found during IdP registration.

External credential short-circuit: when a credential's provision mode is external, the SDK skips RegisterClient during both provisioning and rotate — the client was registered outside the SDK and should not be touched. New constant CredProvisionModeExternal introduced.

GetProvisionMode() added to the CredentialRequest interface and its mock to expose the provision mode to provisioning handlers.

SetIDPResourceName / GetIDPResourceName added to both the provider registry and IdP registry interfaces and implementations, storing the Engage resource name keyed by metadata URL for later lookup.

Api-server models regenerated from branch APIGOV-32071: new IdentityProviderMetadata resource, IdentityProvider policies subresource, Credential.spec.provision.mode field, and updated IdentityProvider spec shape.


pkg/agent/idplifecycle.go
```diff
@@ -0,0 +1,156 @@
+package agent
```
vivekschauhan, line 1: Changes are good for IdP managed/registered by agent configuration, however we might need to expose the life cycle to internally managed OAuth, for e.g. v7 agent needs to be able to create IdP resource(after lookups) based on the metadata that will be provided by agent(not fetched from IdP).

pkg/agent/idplifecycle.go
```diff
@@ -0,0 +1,156 @@
...
+	idpLogger.WithField("name", idpResource.Name).Info("credential policies applied to IdentityProvider")
+}
+
+func createIDPMetadataResource(idpLogger log.FieldLogger, idp config.IDPConfig, idpName string) {
+	idpLogger.WithField("name", idpName).Debug("creating IdentityProviderMetadata resource")
+
+	httpClient := api.NewClient(agent.cfg.GetTLSConfig(), agent.cfg.GetProxyURL())
+	idpLogger.WithField("metadataUrl", idp.GetMetadataURL()).Trace("fetching IdP server metadata")
+
+	serverMetadata, err := oauth.FetchMetadata(httpClient, idp.GetMetadataURL())
```
vivekschauhan: Not sure if we want to refetch the metadata, the calls will be triggered on idp getting registered by agent configuration that will already fetch the metadata with GetAuthProviderRegistry().RegisterProvider()

sbolosan: Good catch —addressed. The earlier implementation called oauth.FetchMetadata directly during metadata resource creation, which was a second HTTP round-trip to the same discovery URL that RegisterProvider had already hit. It now uses provider.GetMetadata() which returns the authServerMetadata struct already cached on the provider during NewProvider. No re-fetch

pkg/agent/idplifecycle.go
sbolosan: Took a stab at this @vivekschauhan .  Let me know if i'm off base. ManageIDPResourceWithMetadata is now the public entry point for agents like v7 that supply metadata directly without a discovery URL. It validates the metadata struct, registers the provider via RegisterProviderWithMetadata (skipping any outbound HTTP fetch), and then runs the same Engage lifecycle — duplicate detection, IdP creation, policy write, metadata resource creation — as the standard config-driven path. No second code path, same lifecycle shared by both.  let me know your thoughts.

sbolosan: Separate v7 cache: ManageIDPResource was incorrectly using ProviderRegistry to cache IdentityProvider resource names. Replaced with a dedicated idpMetadataResourceCache (map[string]string + sync.RWMutex) keyed by TokenEndpoint. External IdP path unchanged.

API query URL: Was incorrectly environment-scoped via GetEnvironmentURL(). Fixed to management.NewIdentityProviderMetadata("", "").GetKindLink() → GET /management/v1alpha1/idpmetadatas

jcollins-axway (dismissed):

pkg/agent/idplifecycle.go
```diff
@@ -0,0 +1,155 @@
...
+		GetAuthProviderRegistry().SetIDPResourceName(idp.GetMetadataURL(), name)
+	}
+	return name
+}
+
+// ManageIDPResource creates or reuses an IdentityProvider resource in Engage using
+// pre-resolved metadata. Public entry point for agents like v7 that supply metadata
+// directly without a discovery URL.
+// Returns the Engage IdentityProvider resource name, or empty string on failure.
+func ManageIDPResource(idpLogger log.FieldLogger, metadata *oauth.AuthorizationServerMetadata) string {
```
sbolosan: @vivekschauhan can you go through this process again to make sure it's all here.

pkg/apic/apiserver/models/catalog/v1alpha1/model_credential_spec.go
```diff
@@ -12,7 +12,8 @@ package catalog
 // CredentialSpec  (catalog.v1alpha1.Credential)
 type CredentialSpec struct {
 	// Reference to Credential Request Definition resource
-	CredentialRequestDefinition string `json:"credentialRequestDefinition"`
+	CredentialRequestDefinition string                  `json:"credentialRequestDefinition"`
+	Provision                   CredentialSpecProvision `json:"provision,omitempty"`
```
vivekschauhan: Should this be a pointer ?

sbolosan: Good catch. You're right — Provision CredentialSpecProvision with omitempty on a value type won't actually omit the field during serialization. Changing it to a pointer. Will also add nil guards at the two direct access sites in credential.go (cred.Spec.Provision.Mode) that are currently relying on zero-value safety.  Will get with sagar so I can regen

sbolosan: nm, its in the generation script, not the yaml file

jcollins-axway (approved):

PR #1037, opened 2026-05-13
APIGOV-32662 - add SetRevisionOnly to ServiceBuilder to skip instance creation for shared-ID proxies

sbolosan:
When API Manager has multiple frontend proxies sharing the same API ID, we were creating a redundant APIServiceInstance for each one. Only one should own the instance.

To fix this, I added SetRevisionOnly() to the service builder so agents can flag a proxy as revision-only, meaning it publishes its spec but skips instance creation. The check is a simple early return inside processInstance. Added a getter IsRevisionOnly() on ServiceBody to keep it consistent with the existing pattern.

### DIFF
--- a/pkg/apic/apiservice.go
+++ b/pkg/apic/apiservice.go
@@ -81,6 +81,16 @@ func (c *ServiceClient) getOwnerObject(serviceBody *ServiceBody, warning bool) (
 	return nil, nil
 }
 
+func (c *ServiceClient) setSpecHashesOnServiceBody(serviceBody *ServiceBody, svc apiv1.Interface) {
+	agentDetails := util.GetAgentDetails(svc)
+	if revDetails, found := agentDetails[specHashes]; found {
+		if specHashes, ok := revDetails.(map[string]interface{}); ok {
+			serviceBody.specHashes = specHashes
+		}
+	}
+
+}
+
 func (c *ServiceClient) updateAPIService(serviceBody *ServiceBody, svc *management.APIService) {
 	owner, ownerErr := c.getOwnerObject(serviceBody, true)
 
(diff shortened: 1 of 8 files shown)


jcollins-axway:
I do not see anything blocking the API Service update if revision only is set. 

If the service does not exist we should create, but we should not update it at all.

jcollins-axway (changes requested):
See comment

pkg/apic/apiservice.go
```diff
@@ -195,9 +195,16 @@ func (c *ServiceClient) processService(serviceBody *ServiceBody) (*management.AP
 
 	if svc != nil {
 		serviceBody.serviceContext.serviceAction = updateAPI
+		c.updateAPIService(serviceBody, svc) // populates from existing service
```
jcollins-axway:
```go
if svc != nil && serviceBody.IsRevisionOnly() {
 serviceBody.serviceContext.serviceName = svc.Name
 serviceBody.serviceContext.serviceID = svc.Metadata.ID
 return svc, nil
} else if svc != nil {
 serviceBody.serviceContext.serviceAction = updateAPI
 httpMethod = http.MethodPut
 serviceURL += "/" + svc.Name
 c.updateAPIService(serviceBody, svc)
} else {
 svc = c.buildAPIService(serviceBody)
}
```

jcollins-axway (dismissed):

vivekschauhan (dismissed):

sbolosan (commented):
Looks good

pkg/apic/apiservicerevision.go
```diff
@@ -146,23 +140,8 @@ func (c *ServiceClient) getRevisions(queryString string) ([]*management.APIServi
...
-		}
-		serviceBody.serviceContext.revisionCount = totalCount
-		apiServiceRevisions = revisions
-	}
-	return apiServiceRevisions, nil
-}
-
 // checkAndUpdateExistingRevision checks if a revision with the same hash exists and updates tags if needed
-func (c *ServiceClient) checkAndUpdateExistingRevision(serviceBody *ServiceBody, apiServiceRevisions []*management.APIServiceRevision) (bool, error) {
+func (c *ServiceClient) checkAndUpdateExistingRevision(serviceBody *ServiceBody) (bool, error) {
```
sbolosan, line 144: update comment to something like "checks if a revision with the same hash was already published."

jcollins-axway (approved):

vivekschauhan (approved):

PR #1040, opened 2026-05-24
APIGOV-31452 - Send New Insights Event Formats

sbolosan:
Delivers the SDK changes needed to produce metric v3 and transaction v2 Insights event formats.

Transaction events now emit the correct versioned InsightsEvent envelope for both leg and summary events, with the right field structure, owner information, and reporter block that analytics expects.

Metric events are updated to v3 — 

- carrying deployment
- runtime type
- owner context for the API, application, and product
- Owner resolution is centralized into a set of shared helpers that cover all the standard cases including the new x-private user ownership type, ready for when the server-side change lands to propagate that data.

The AccessRequest cache loading was updated so owner data from embedded references is always present at startup and on watch events, ensuring agents have what they need to populate owner fields correctly.


vivekschauhan (dismissed):

jcollins-axway:
we should hold this until we get the last piece of the puzzle in the event

pkg/traceability/httpclient.go
```diff
@@ -178,13 +178,13 @@ func (client *HTTPClient) publishEvents(data []publisher.Event) error {
 			timeStamp = event.Content.Timestamp
 			allFields, err := event.Content.Fields.GetValue("fields")
 			if err != nil {
-				client.headers[FlowHeader] = TransactionFlow
+				delete(client.headers, FlowHeader)
```
vivekschauhan, line 181: If I understand correctly, axway-target-flow header will only be included for metric events, hope we have confirmed this with analytics team

vivekschauhan (dismissed):

pkg/traceability/httpclient.go
sbolosan: Yup, that's correct.  Alasdair, confirmed.

pkg/agent/handler/credential_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/Axway/agent-sdk/pkg/watchmanager/proto"
 )
 
+const algRSAOAEP = "RSA-OAEP"
```
jcollins-axway: can we use the existing constant in util package?

pkg/agent/handler/traceaccessrequest.go
```diff
@@ -48,7 +48,11 @@ func (h *traceAccessRequestHandler) Handle(ctx context.Context, _ *proto.EventMe
 	if h.shouldProcessForTrace(ar.Status, ar.Metadata.State) {
 		cachedAccessReq := h.cache.GetAccessRequest(resource.Metadata.ID)
 		if cachedAccessReq == nil {
-			h.cache.AddAccessRequest(resource)
+			enriched, err := h.client.GetResource(resource.GetSelfLink() + "?embed=metadata.references")
+			if err != nil || enriched == nil {
+				enriched = resource
+			}
+			h.cache.AddAccessRequest(enriched)
```
jcollins-axway, line 62: is there a case that the cache, maybe persisted, doe snot have the enriched data? Should we check if the cached access request is enriched before determining if its a noop?

pkg/agent/handler/traceaccessrequest_test.go
```diff
@@ -1,119 +1,166 @@
...
-	err := handler.Handle(NewEventContext(proto.Event_CREATED, nil, ri.Kind, ri.Name), nil, ri)
-	assert.Nil(t, err)
-}
-
-func TestTraceAccessRequestTraceHandler(t *testing.T) {
-	cm := agentcache.NewAgentCacheManager(&config.CentralConfiguration{}, false)
-	c := &mockClient{}
-	handler := NewTraceAccessRequestHandler(cm, c)
+func makeTraceAR(id, name, appName, instanceID string) *management.AccessRequest {
 	ar := &management.AccessRequest{
```
jcollins-axway: can we return here instead of creating the ar, adding the status, then returning ar?

pkg/agent/discoverycache.go
```diff
@@ -246,7 +246,14 @@ func (dc *discoveryCache) buildResourceFunc(filter management.WatchTopicSpecFilt
 		logger := dc.logger.WithField("kind", filter.Kind)
 		logger.Tracef("fetching %s and updating cache", filter.Kind)
 
-		resources, err := dc.client.GetAPIV1ResourceInstances(nil, ri.GetKindLink())
+		var queryParams map[string]string
+		if filter.Kind == management.AccessRequestGVK().Kind {
+			queryParams = map[string]string{"embed": "metadata.references"}
+		} else if filter.Kind == management.APIServiceGVK().Kind {
+			queryParams = map[string]string{"embed": "x-agent-details"}
```
jcollins-axway: is this needed? x-agent-details is already added to the resource

pkg/apic/apiserver/models/api/v1/resourcemeta_test.go
```diff
@@ -9,6 +9,13 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	xAgentDetails  = "x-agent-details"
```
jcollins-axway: We should already have a constant for this in the definitions package

pkg/transaction/models/definitions.go
```diff
@@ -175,3 +179,11 @@ type ObservationDetails struct {
 	Start int64 `json:"start,omitempty"`
 	End   int64 `json:"end,omitempty"`
 }
+
+// OwnerBlock represents the owner of an API or application resource in insights events.
+// Type is one of "team", "user", "none", or "unknown".
+type OwnerBlock struct {
```
jcollins-axway: Could this just be Owner and not OwnerBlock? Not sure I like the work "Block"

pkg/transaction/util/ownerresolver_test.go
```diff
@@ -0,0 +1,170 @@
...
+	management "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/management/v1alpha1"
+	"github.com/Axway/agent-sdk/pkg/config"
+	"github.com/Axway/agent-sdk/pkg/transaction/models"
+)
+
+const (
+	ownerNilReturnsNone = "nil owner returns none"
+	ownerTeamWithGUID   = "team owner with GUID returns team block"
+	ownerTeamEmptyGUID  = "team owner with empty GUID returns unknown"
+	ownerUserXPrivate   = "x-private team owner returns user block"
```
jcollins-axway: do not see the point of making a constant out of a test name, maybe the test name should be more specific? Does sonar care about _test files?

pkg/transaction/eventgenerator_test.go
```diff
@@ -119,13 +140,217 @@ func TestCreateEventWithInvalidTokenRequest(t *testing.T) {
...
+				TransactionID:      "txn-beat-sum",
+				TransactionSummary: &Summary{Status: "Success"},
+			},
+			wantEvent: "api.transaction.summary",
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			cfg := createMapperTestConfig(s.URL, tc.tenantID, "prod", "env-name", tc.envID)
```
jcollins-axway, line 215: rather than specifying the default tenant and env ids for each test could we set them to their default prior to calling this? I guess it is 6 in 1 half doze in the other.

pkg/transaction/v2event.go
```diff
@@ -0,0 +1,534 @@
...
+}
+
+// insightsProxy is the nested proxy sub-object for leg and summary v2 data.
+type insightsProxy struct {
+	ID   string `json:"id,omitempty"`
+	Name string `json:"name,omitempty"`
+}
+
+// TransactionLegV2Data is the data payload for api.transaction.event (version "2").
+type TransactionLegV2Data struct {
```
jcollins-axway: Are we still using the basic TransactionLeg structure that this needs V2?

pkg/transaction/v2event.go
```diff
@@ -0,0 +1,534 @@
...
+	Host   string `json:"host,omitempty"`
+}
+
+// insightsResourceRef is a minimal {id} sub-object.
+type insightsResourceRef struct {
+	ID string `json:"id,omitempty"`
+}
+
+// TransactionSummaryV2Data is the data payload for api.transaction.summary (version "2").
+type TransactionSummaryV2Data struct {
```
jcollins-axway: Same question

pkg/agent/handler/traceaccessrequest.go
sbolosan: yes, you're right. Adding code to check

pkg/agent/discoverycache.go
sbolosan: residual code on my part, it is already on the resource. removing

pkg/transaction/util/ownerresolver_test.go
jcollins-axway: this was marked as resolved but no changes or reply?

pkg/transaction/eventgenerator_test.go
jcollins-axway: this was marked as resolved but no changes or reply?

pkg/transaction/v2event.go
jcollins-axway: this was marked as resolved but no changes or reply?

pkg/transaction/v2event.go
jcollins-axway: this was marked as resolved but no changes or reply?

pkg/transaction/util/ownerresolver_test.go
jcollins-axway: nevermind, had the wrong changelist

pkg/transaction/v2event.go
jcollins-axway: same, wrong changelist

pkg/transaction/v2event.go
jcollins-axway: same, wrong changelist

jcollins-axway (dismissed):

jcollins-axway (dismissed):

vivekschauhan (dismissed):

vivekschauhan (approved):

PR #1048, opened 2026-06-15
APIGOV-31191 - Cache validation/rebuild refactoring

vivekschauhan:
- Remove API calls with metadata for cache validation check
- Remove entire cache flush. Flush only the kind getting processed and after the API call is successful to avoid cache inconsistency
- Cleanup instance count map that should speed up building APIservice cache items
- Keep original APIService resource instance in cache when duplicates are added to the cache
- Hook cache rebuild on stream client reconnection instead of harvester error
- Check publishing lock and acquire lock while publishing


pkg/agent/cache/manager.go
```diff
@@ -266,8 +260,6 @@ func (c *cacheManager) setLoadedCache(lc cache.Cache, key string) {
 		c.teams = lc
 	case managedAppKey:
 		c.managedApplicationMap = lc
-	case subscriptionsKey:
-		c.subscriptionMap = lc
```
jcollins-axway, line 270: good call

pkg/agent/cache/manager.go
```diff
@@ -154,13 +153,11 @@ type cacheManager struct {
 	jobs.Job
 	logger                  log.FieldLogger
 	apiMap                  cache.Cache
-	instanceCountMap        cache.Cache
```
jcollins-axway, line 157: aside from moving, any migration we have to do? probably not when a rebuild does happen it will take care of it?

pkg/agent/resource/manager.go
```diff
@@ -166,7 +167,7 @@ func (a *agentResourceManager) UpdateAgentStatus(status, prevStatus, message str
 
 	// See if we need to rebuildCache
 	if needsRebuild, filters := a.shouldRebuildCache(); needsRebuild && a.rebuildCache != nil {
-		a.rebuildCache.RebuildCache(filters...)
+		a.rebuildCache.RebuildCache(context.Background(), filters...)
```
jcollins-axway: Should we start with a consistent context, maybe create one on agentResourceManager? Perhaps it could be used in other places, now or future?

pkg/agent/discovery.go
jcollins-axway: This is not new, I realize, but if agent.apiClient is nil would we hit an NPE here. 

Maybe our check at line 135 should check if nil and return err?

pkg/agent/discovery_test.go
```diff
@@ -0,0 +1,88 @@
...
+	// PublishAPI must not deadlock — it skips re-acquiring the already-held lock
+	err := PublishAPI(apic.ServiceBody{})
+	assert.Nil(t, err)
+	assert.True(t, publishCalled)
+	// flag remains true because the caller still holds the lock
+	assert.True(t, agent.publishingLockAcquired.Load())
+
+	PublishingUnlock()
+	assert.False(t, agent.publishingLockAcquired.Load())
+}
```
jcollins-axway, line 92: Any way to combine these 2 test funcs into a single func with multiple cases? I think that is the format of tests we tend to favor.

Of course, if that makes sense here.

pkg/agent/cache/manager.go
vivekschauhan: Not needed as the next save will store the desired cache items

pkg/agent/discovery.go
vivekschauhan: Done

pkg/agent/discovery_test.go
vivekschauhan: Done

pkg/agent/resource/manager.go
vivekschauhan: Agree, we should but may be in a separate change

jcollins-axway (approved):

PR #1049, opened 2026-06-16
APIGOV-32943 — Okta IDP per-scope policy lifecycle, app name templates, and scope exclude list

sbolosan:
Automated policy management :
When a credential is provisioned, the agent now creates or updates an Okta authorization policy scoped to the specific access being granted. When a credential is deprovisioned, the agent removes the credential from that policy. If it was the last credential on the policy, the agent automatically cleans up the policy entirely rather than leaving empty policies behind in Okta.

Configurable naming :
Okta app and policy names can now be built from a configurable template using placeholders for the Marketplace application name, owning team, credential name, scope, and OAuth flow. Names are validated at startup and length limits are enforced before anything is sent to Okta.

Team name resolution :
When registering an app in Okta, if the owning team isn't immediately available in the local cache, the agent will look it up from the platform and cache it for future use.

Scope filtering : 
Scopes returned by Okta (such as openid, profile, email) can be filtered out from what is shown in the Marketplace UI using a configurable exclude list.


pkg/agent/handler/credential.go
```diff
@@ -218,43 +218,52 @@ func (h *credentials) onDeleting(ctx context.Context, cred *management.Credentia
...
+		clientID := provCreds.idpProvisioner.GetIDPCredentialData().GetClientID()
+		tokenURL := provCreds.GetIDPProvider().GetTokenEndpoint()
+		providerName := provCreds.GetIDPProvider().GetName()
+		logger = logger.WithField("client_id", clientID).WithField("provider", providerName).WithField("tokenURL", tokenURL)
+
+		unregisterErr := provCreds.idpProvisioner.UnregisterClient()
+		if unregisterErr != nil {
+			logger.WithError(unregisterErr).Warn("error deprovisioning credential request from IDP, please ask administrator to remove the client from IdP")
+		} else {
+			if err := removeClientIDFromManagedApp(logger, h.client, app, clientID, tokenURL); err != nil {
```
vivekschauhan, line 244: Can we add a check for okta idp similar to the one when registering the client
```
provCreds.GetIDPProvider().GetConfig().GetIDPType() != oauth.TypeOkta
```

dgghinea: it's not exactly needed because during provisioning for non-okta, the clientID is not added to the x-agent-details.clientIDs array

jcollins-axway (dismissed):

vivekschauhan (dismissed):

alrosca (dismissed):

jcollins-axway (dismissed):

vivekschauhan (dismissed):

vivekschauhan (approved):

PR #1055, opened 2026-06-29
another attempt to fix concurrent metric collector test

dgghinea:
Maybe 70000 goroutines in a single test was a bit too much

### DIFF
--- a/pkg/transaction/metric/metricscollector_test.go
+++ b/pkg/transaction/metric/metricscollector_test.go
@@ -543,26 +543,10 @@ func TestConcurrentMetricCollectorEvents(t *testing.T) {
 	apiDetails := []models.APIDetails{
 		{ID: "000", Name: "000", Revision: 1, TeamID: teamID},
 		{ID: "111", Name: "111", Revision: 1, TeamID: teamID},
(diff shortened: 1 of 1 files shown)


jcollins-axway:
By removing this we are reducing the number of concurrent metrics being given to the SDK, I think by removing this we are invalidating the purpose of the test.

jcollins-axway (changes requested):
can not see the benefit of reducing this invalidating the reason for the test

dgghinea:
@jcollins-axway it's not exactly removing the test, it's reducing its usage. I think the test is still valid but it's testing 2800 concurrent calls instead of 70000.
I initially thought this was the reason the GitLab runner was failing. but if you think this is not related, I can just close this.

jcollins-axway:
@dgghinea is it helping anything with the reduction?

dgghinea:
> @dgghinea is it helping anything with the reduction?

I took a look and the time of the test pipeline to finish went from around ~9 mins to 3.5 mins(specifically, the test that was modified went from ~300s to 4s). So I'd say yes

jcollins-axway (approved):

PR #1056, opened 2026-06-30
APIGOV-33097 - buffer the events channel to allow logging receipt of events

jcollins-axway:


sbolosan:
code looks good. Maybe an updated unit test for logEventReceipt

jcollins-axway:
> code looks good. Maybe an updated unit test for logEventReceipt

do we really think that is necessary?

I ask because there would be no assertion that could be made with a unit test for that, it would only be for line coverage. Either the function logs something, which we do not assert on, and returns nothing. Or it does not log anything and returns nothing.

dgghinea (approved):

vivekschauhan (approved):

alrosca (approved):

sbolosan (approved):

PR #1059, opened 2026-07-13
APIGOV-33163 - new event kind handler structure

dgghinea:


pkg/agent/handler/accessrequest.go
```diff
@@ -58,13 +58,27 @@ func NewAccessRequestHandler(prov prov.AccessProvisioner, cache agentcache.Manag
...
-func (h *accessRequestHandler) Handle(ctx context.Context, meta *proto.EventMeta, resource *apiv1.ResourceInstance) error {
+func (h *accessRequestHandler) Kinds() []string {
+	return []string{management.AccessRequestGVK().Kind}
+}
+
+func (h *accessRequestHandler) ShouldHandle(ctx context.Context, event *proto.Event) bool {
 	action := GetActionFromContext(ctx)
-	if resource.Kind != management.AccessRequestGVK().Kind {
-		return nil
+	if event.Payload.Kind != management.AccessRequestGVK().Kind {
```
vivekschauhan: Do we still need this check with handlers invoked for the kind?

dgghinea: oh, I see what you're saying. With the new map approach, no. These can be removed

pkg/agent/discoverycache.go
sbolosan: looks like this is dead code now

pkg/agent/cachevalidationjob.go
```diff
@@ -128,7 +128,9 @@ func (cv *cacheValidator) sequenceInSync() bool {
 	}
 
 	cachedSeq := cv.sequence.GetSequence()
-	if serverSeq != cachedSeq {
+	// At start-up or agent restart, because of the status change event, the serverSeq will always be higher with 1 than the cachedSeq.
```
sbolosan: maybe just a comment thing for me.  looks like this code is also called from validateAndRebuildCache. So this +1 tolerance is also processing every mid run reconnect as well, no?

pkg/agent/handler/discoverymanagedapplication.go
```diff
@@ -21,13 +21,24 @@ func NewDiscoveryManagedApplicationHandler(cache agentcache.Manager) Handler {
 	}
 }
 
+func (h *discoveryManagedApplication) ShouldHandle(ctx context.Context, event *proto.Event) bool {
```
sbolosan: Don't know if i'm seeing this correctly.  Could you take another look at the ShouldHandle here (and the same pattern in discoveryaccessrequest.go:24, tracemanagedapplication.go:24, and traceaccessrequest.go:27)? 

It returns false once h.cache.GetManagedApplication(event.Payload.Metadata.Id) finds an existing entry. But what about the DELETE action. Since Handle() never runs when ShouldHandle() returns false, it looks like once a resource is cached, later update and delete events for that same ID might not get processed by this handler anymore??? That's what i'm seeing. Just checking in case it's worth a second pass.. could be intentional, but wanted to double check it.

pkg/agent/handler/agentresource.go
```diff
@@ -72,27 +72,50 @@ func NewAgentResourceHandler(agentResourceManager resource.Manager, sampler samp
...
+	if event.Type == proto.Event_SUBRESOURCEUPDATED &&
+		event.Metadata.Subresource != definitions.XAgentDetails &&
+		!(event.Payload.Kind == management.TraceabilityAgentGVK().Kind && event.Metadata.Subresource == management.TraceabilityAgentAgentstateSubResourceName) {
+		return false
+	}
+	if event.Type != proto.Event_UPDATED && event.Type != proto.Event_SUBRESOURCEUPDATED {
+		return false
 	}
+
+	return true
```
jcollins-axway: Maybe it is just me, but it is somewhat mind boggling to read these conditionals and know when we are handling the event. Can we simplify the conditionals at all?

```go
if event.Type == proto.Event_CREATED {
  return false
}
if event.Type == proto.Event_DELETED {
  return false
}
if event.Type == proto.Event_UPDATED {
  return true
}

// Sub resource update
if event.Metadata.Subresource == definitions.XAgentDetails {
  return true
}
if event.Payload.Kind == management.TraceabilityAgentGVK().Kind && event.Metadata.Subresource == management.TraceabilityAgentAgentstateSubResourceName {
  return true
}

return false
```

Maybe we can simplify other should handle funcs?

jcollins-axway (dismissed):

pkg/agent/cachevalidationjob.go
dgghinea: I think mid run reconnects would happen in case the watch controller context gets cancelled(SDK stream client timeout), which would mark the agent as stopped and still trigger a status change to Stopped

pkg/agent/handler/discoverymanagedapplication.go
dgghinea: hm... I was trying to not change the old behaviour for this code. I think we're not interested in any updates apart from the resource itself being there.
Regarding your statement for the DELETE action, you're right. good call

jcollins-axway (dismissed):

pkg/agent/cachevalidationjob.go
vivekschauhan: I think this +1 check is not needed. Watch-controller agent status update will not raise events, so the sequence should be same

dgghinea: hm... weird. I was permanently getting sequence mismatches(+1 difference) during testing and my assumption was that this was the reason. I guess I'll revert it then

jcollins-axway (dismissed):

sbolosan (approved):

jcollins-axway (approved):

PR #1060, opened 2026-07-14
APIGOV-33179 - Replace v1alpha1 api version with v1

vivekschauhan:


pkg/agent/aclupdatejob_test.go
```diff
@@ -123,7 +123,7 @@ func TestACLUpdateHandlerJob(t *testing.T) {
 					token := "{\"access_token\":\"somevalue\",\"expires_in\": 12235677}"
 					resp.Write([]byte(token))
 				}
-				if strings.Contains(req.RequestURI, "/apis/management/v1alpha1/environments/"+test.envName+"/accesscontrollists") {
+				if strings.Contains(req.RequestURI, "/apis/management/v1/environments/"+test.envName+"/accesscontrollists") {
```
jcollins-axway: not that it needs to be updated here, but probably should just be using the KindLink function for this, no?

jcollins-axway (dismissed):

pkg/agent/aclupdatejob_test.go
vivekschauhan: Done

dgghinea (approved):

PR #1068, opened 2026-07-27
APIGOV-33168 - use only required fields for querying API Server

dgghinea:


pkg/agent/events/eventlistener.go
jcollins-axway: should we use the new logger var here too?

dgghinea: oh, you're right. thanks

jcollins-axway (approved):

sbolosan (approved):

PR #1071, opened 2026-07-30
APIGOV-33165 - concurrency event listener

dgghinea:


pkg/agent/events/eventlistener.go
```diff
@@ -141,15 +204,77 @@ func (em *EventListener) handleEvent(event *proto.Event) error {
...
+}
+
+// managedApplicationKey returns the ManagedApplication name an event relates to, for the three
+// kinds that participate in managed-app provisioning. ok is false for any other kind, or when
+// the reference can't be determined from ri (e.g. AccessRequest/Credential delete events, whose
+// spec is empty).
+func managedApplicationKey(event *proto.Event, ri *apiv1.ResourceInstance) (key string, ok bool) {
+	switch event.Payload.Kind {
+	case management.ManagedApplicationGVK().Kind:
+		return event.Payload.Name, true
```
jcollins-axway, line 342: I know name is immutable but I think using the id would be better, thoughts?

pkg/agent/events/eventlistener.go
```diff
@@ -141,15 +204,77 @@ func (em *EventListener) handleEvent(event *proto.Event) error {
...
+	case management.AccessRequestGVK().Kind, management.CredentialGVK().Kind:
+		if ri == nil || ri.Spec == nil {
+			return "", false
+		}
+		key, ok = ri.Spec["managedApplication"].(string)
```
jcollins-axway, line 347: I guess if we use the man app id then we would have to look in references here, so if that is not possible then never mind on the id

pkg/agent/events/eventlistener.go
```diff
@@ -39,41 +49,82 @@ type EventListener struct {
...
+	return func(el *EventListener) {
+		el.ctx = ctx
+		el.cancel = cancel
+	}
+}
 
 // NewEventListener creates a new EventListener to process events based on the provided Handlers,
 // indexed by the resource Kind they should be dispatched for.
-func NewEventListener(ctx context.Context, cancel context.CancelCauseFunc, source chan *proto.Event, client APIClient, baseURL string, sequenceManager SequenceProvider, handlersByKind map[string][]handler.Handler) *EventListener {
+func NewEventListener(source chan *proto.Event, client APIClient, baseURL string, sequenceManager SequenceProvider, handlersByKind map[string][]handler.Handler, opts ...ListenerOpts) *EventListener {
```
jcollins-axway: was there a reason to make this change where the context and cancel are optional?

dgghinea: half. I initially had changes with a configurable number of workers and thought this func has too many params, so I thought about adding functional options and included the ctx + cancel since those + the nbOfWorkers are really the ones that are "optional". I can convert them back.

pkg/agent/events/eventlistener.go
dgghinea: yes, that was mostly the reason for not using the ID. Wanted to avoid having to look up in the references.
I suppose when a mismatch(incorrectly putting it in the low prio queue) would happen is for managedApps with the same name from different marketplaces? which I don't think should be problematic

pkg/agent/events/eventlistener.go
jcollins-axway: I am thinking cancel could be optional, but context should not be. imo

pkg/agent/events/eventlistener.go
```diff
@@ -6,21 +6,31 @@ import (
...
 	"github.com/Axway/agent-sdk/pkg/util/log"
 
 	apiv1 "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/api/v1"
+	management "github.com/Axway/agent-sdk/pkg/apic/apiserver/models/management/v1"
 	"github.com/Axway/agent-sdk/pkg/watchmanager/proto"
 )
 
-// Listener starts the EventListener
+const (
+	lowQueue     = "lowPriorityQueue"
```
jcollins-axway: This is just a provisioningQueue, right? Would it be better to name it like that?

pkg/agent/events/eventlistener.go
```diff
@@ -39,41 +55,142 @@ type EventListener struct {
 	logger          log.FieldLogger
 	sequenceManager SequenceProvider
 	source          chan *proto.Event
+	kindJobs        map[string]chan handlerData // one worker per kind
+	lowPriorityJobs chan handlerData            // exactly one dedicated worker
```
jcollins-axway: provisioningJobs?

pkg/agent/events/eventlistener.go
dgghinea: understood. will revert

pkg/agent/events/eventlistener.go
dgghinea: yeah, I had this name in mind from League of Legends and I thought "this is a cool opportunity to use it"

pkg/agent/discoverycache.go
```diff
@@ -300,18 +301,30 @@ func (dc *discoveryCache) handleResourcesList(list []*apiv1.ResourceInstance) er
...
-			return err
+		if err := h.Handle(ctx, nil, ri); err != nil {
+			logger.WithError(err).Error("failed to handle discovery cache resource")
 		}
 	}
 	return nil
 }
 
+func getAction(state, kind string) proto.Event_Type {
+	// MPResource Handlers(non-cache ones) don't react anymore to to a Event_CREATED.
```
sbolosan: double to to in comment ;)

pkg/agent/discoverycache.go
```diff
@@ -300,18 +301,30 @@ func (dc *discoveryCache) handleResourcesList(list []*apiv1.ResourceInstance) er
...
-		if !ok {
-			continue
+		if ch, ok := h.(handler.CacheHandler); ok {
+			if err := ch.HandleCache(ri); err != nil {
+				logger.WithError(err).Error("failed to handle discovery cache resource")
+			}
 		}
-		if err := ch.HandleCache(ri); err != nil {
-			return err
+		if err := h.Handle(ctx, nil, ri); err != nil {
```
sbolosan, line 317: if I'm looking at this right, h.Handle(ctx, nil, ri) now runs for every handler during cache rebuild, skipping ShouldHandle() entirely. Oh and actually the h.prov == nil check as well.

accessrequest.go used to guard against nil prov inside Handle() too, but it looks like that got removed. Do you believe h.prov can ever actually be nil here, or is that guard safe to drop?

dgghinea: ShouldHandle() is useful when we don't want to fetch the resource. in the discoverycache part, the resource is already fetched, so there's no need for that call.
But I think I mixed up some cache deletions operations when a delete event comes in, so I think it's a good mention and I'll put that back. Good call :D.

prov won't probably ever be nil, but by adding the ShouldHandle() back, it should do the check there.

PR #1072, opened 2026-07-31
APIGOV-33344 - update trans leg and summary with only required fields

sbolosan:
Fix:

Remove the following out-of-spec fields specifically from api.transaction.event (DO NOT remove from api.transaction.summary *or *api.transaction.status.metric)

data.transactionId
data.legId
data.api
Whole object including data.api.id, data.api.owner.type, data.api.owner.team_guid etc...
data.proxy
Whole object including data.proxy.id, data.proxy.name etc..

Remove the following out-of-spec fields specifically from api.transaction.summary (DO NOT remove from api.transaction.event or api.transaction.status.metric)

data.product.name
data.product.versionName
data.product.owner
Whole object including data.product.owner.type etc...
data.consumerDetails.application.name
data.consumerDetails.publishedProduct.name
data.consumerDetails.subscription.name

### DIFF
--- a/pkg/transaction/v2event.go
+++ b/pkg/transaction/v2event.go
@@ -43,21 +43,18 @@ type insightsAPIDetail struct {
 // insightsConsumerAppDetail is the application sub-object within consumerDetails.
 type insightsConsumerAppDetail struct {
 	ID            string        `json:"id,omitempty"`
-	Name          string        `json:"-"`
 	ConsumerOrgID string        `json:"consumerOrgId,omitempty"`
 	Owner         *models.Owner `json:"owner,omitempty"`
 }
 
 // insightsPublishedProduct is the publishedProduct sub-object within consumerDetails.
 type insightsPublishedProduct struct {
-	ID   string `json:"id,omitempty"`
-	Name string `json:"-"`
+	ID string `json:"id,omitempty"`
 }
 
 // insightsSubscription is the subscription sub-object within consumerDetails.
 type insightsSubscription struct {
-	ID   string `json:"id,omitempty"`
-	Name string `json:"-"`
+	ID string `json:"id,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


jcollins-axway:
Do we utilize those fields within the code still? Just wondering if they need to be in memory ever if we are not going to serialize and send to analytics.

sbolosan:
> Do we utilize those fields within the code still? Just wondering if they need to be in memory ever if we are not going to serialize and send to analytics.

@jcollins-axway 
Good call. Yeah, for the most part. The Proxy field set but never read, so I'll kill those. There is a SetLegProxy call in agents-controller that becomes dead code. So i'll update that

TransactionID and API on TransactionLegData are staying as json:"-". TransactionID is read by GetEventID() which the event pipeline uses to track events, and API is read by the agents-controller for owner enrichment after the leg is built. 

I'll update

pkg/transaction/v2event.go
```diff
@@ -478,31 +453,15 @@ func buildConsumerDetails(cd *models.ConsumerDetails, appOwner *models.Owner) *i
...
+		out.Subscription = &insightsSubscription{ID: cd.Subscription.ID}
 	}
 
 	return out
 }
 
-// SetLegProxy sets the proxy block on a TransactionLegData after it has been built.
-// Used by the agents-controller to populate proxy from enriched API details post-build,
-// since insightsProxy is unexported and cannot be constructed externally.
-func SetLegProxy(leg *TransactionLegData, proxyID, proxyName string) {
```
jcollins-axway, line 499: We just need to make sure no agents are using this setter. Or we can keep it as a noop.

jcollins-axway: Seems you already had, my bad

jcollins-axway (approved):

vivekschauhan (approved):

PR #1077, opened 2026-08-06
APIGOV-33165 - use configurable worker vars

dgghinea:


pkg/agent/events/eventlistener.go
```diff
@@ -17,13 +19,13 @@ import (
...
 
 const (
-	provisioningQueue = "provisioningQueue"
-	regularQueue      = "regularQueue"
-
-	// laneBufferSize lets a lane's dispatch send get a few events ahead of its worker, so a
-	// short burst of slow handling on one lane doesn't immediately stall dispatch for every
-	// other lane - there's only one dispatch goroutine feeding all of them.
-	laneBufferSize = 5
+	regularQueue     = "regularQueue"
```
jcollins-axway, line 22: rather than a string maybe we can make this an iota?

pkg/agent/events/eventlistener.go
```diff
@@ -300,8 +341,12 @@ func (em *EventListener) handleEvent(event *proto.Event) error {
 		}
 	}
 
+	// TODO: right now, we have no delete events because the WatchTopic is not configured to listen for delete calls
+	// for ManagedApp/AccessRequest/Credential. maybe simplest way to fix this is to have a separate goroutine that
+	// checks in the cache(every 10 mins, let's say) for the existing managedAppNames and removes them if not found
```
jcollins-axway, line 346: we should just sub to delete events

jcollins-axway (approved):

pkg/agent/events/eventlistener.go
dgghinea: was under the impression that we wanted to avoid that, but if not, then I guess we can do that

pkg/agent/events/eventlistener.go
dgghinea: it's the only string in here though, not sure if that applies here

jcollins-axway (approved):

PR #1078, opened 2026-08-07
APIGOV-28603 - TA to use updated https endpoints and beats lib drop

sbolosan:
Live confluence doc for
1. changes to sdk, Traceability agents
2. what the equivalent code beats did
3. why the change was needed in sdk
4. what changes are needed in agents after sdk bump and why

https://confluence.axway.com/spaces/RDAPI/pages/1894374460/Update+the+traceability+agents+to+only+support+events+over+HTTP+and+lib+beats+removal

1. Removed the elastic/beats (libbeat) dependency from agent-sdk's traceability publishing path, replacing it with agent-sdk's own lightweight event/client/batch implementation.
2. Traceability agents now publish over a hardened HTTPS endpoint by default instead of the old TCP-based ingestion endpoint.
3. Added startup validation that hard-fails with a clear error on outdated configs (old tcp protocol, port 5044, or ingestion.* hosts), so a misconfigured agent fails fast instead of quietly running against a decommissioned setup.
4. Added a deprecation warning for an old config field that's no longer needed, without breaking startup.
5. Rebuilt the retry/backoff and multi-host failover logic that used to come from beats, so agents keep that resilience without the beats dependency.
6. Rebuilt TLS/certificate handling that used to come from beats, including warning (rather than failing) on now-unsupported settings.
7. Added a reusable event-batch helper so agents publishing traceability data no longer need to write their own boilerplate for it.
8. Simplified the logging configuration code that previously had to bridge to beats' logging system.
9. Added/updated test coverage across all of the above.


go.mod
```diff
@@ -34,70 +33,38 @@ require (
...
-	golang.org/x/telemetry v0.0.0-20260625142307-59b4966ccb57 // indirect
-	golang.org/x/tools v0.47.0 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260414002931-afd174a4e478 // indirect
 	gopkg.in/ini.v1 v1.66.6 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	howett.net/plist v1.0.1 // indirect
 )
 
 replace github.com/Shopify/sarama => github.com/elastic/sarama v1.19.1-0.20210823122811-11c3ef800752
```
jcollins-axway: this replace still needed?

go.mod
```diff
@@ -34,70 +33,38 @@ require (
...
-	github.com/elastic/go-sysinfo v1.15.3 // indirect
-	github.com/elastic/go-ucfg v0.8.8 // indirect
-	github.com/elastic/go-windows v1.0.2 // indirect
-	github.com/fatih/color v1.18.0 // indirect
 	github.com/go-openapi/jsonpointer v0.22.5 // indirect
 	github.com/go-openapi/swag/jsonname v0.25.5 // indirect
 	github.com/goccy/go-json v0.10.2 // indirect
-	github.com/gofrs/uuid v4.4.0+incompatible // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
-	github.com/google/pprof v0.0.0-20250630185457-6e76a2b096b5 // indirect
```
jcollins-axway, line 53: may want to make sure pprof still works for our agents

go.mod
sbolosan: you're right, not needed any more.

go.mod
sbolosan: Confirmed working. Started httpprof=true and hit the live endpoints. Endpoints work exactly the same with the module gone. So it looks like actual pprof endpoints (/debug/pprof/* in healthchecker.go, CPU/heap profiling in agent.go) are built on net/http/pprof/runtime/pprof. So removing it doesn't touch our profiling capability.

pkg/cmd/properties/properties.go
```diff
@@ -441,6 +444,18 @@ func (p *properties) parseStringValueForKey(key string) string {
 	return s
 }
 
+func (p *properties) parseStringValue(key string) string {
```
jcollins-axway, line 447: why move it up 20 lines? not that it matters just interesting

pkg/cmd/properties/properties.go
```diff
@@ -122,8 +123,20 @@ func WithUpperLimitInt(upper int) IntOpt {
 	}
 }
 
+// aliasKeyPrefix - when set, properties are looked up under "<aliasKeyPrefix>.<key>" first,
+// falling back to the plain "<key>".
 var aliasKeyPrefix string
 
+// SetAliasKeyPrefix -
```
jcollins-axway, line 130: This, I like that it was moved

pkg/cmd/root.go
jcollins-axway, line 110: Not that it needs done now, but I thought we would be able to remove one of these command creations. Maybe a backlog item?

pkg/cmd/properties/properties.go
sbolosan: I took this out originally. That's what broke v7 traceability for yaml/environments vars. I guess when i cherry picked it, it moved it there?

pkg/cmd/root.go
sbolosan: I have time to update this. 👍🏾

pkg/traceability/backoff.go
```diff
@@ -0,0 +1,80 @@
+package traceability
```
jcollins-axway, line 1: Another backlog thing, it would be great to throw this along side our api client and be able to use it for all api clients.

pkg/traceability/errors.go
```diff
@@ -4,8 +4,13 @@ import "github.com/Axway/agent-sdk/pkg/util/errors"
...
-	ErrInvalidRegex     = errors.Newf(1506, "could not compile the %s regex value (%v): %v")
+	ErrHTTPNotConnected       = errors.New(1503, "http transport is not connected")
+	ErrJSONEncodeFailed       = errors.New(1504, "failed to encode the json content")
+	ErrInvalidConfig          = errors.Newf(1505, "invalid traceability config. Config error: %s")
+	ErrInvalidRegex           = errors.Newf(1506, "could not compile the %s regex value (%v): %v")
+	ErrTCPProtocolRemoved     = errors.New(1507, "protocol 'tcp' is no longer supported; set output.traceability.protocol to https")
+	ErrPort5044Removed        = errors.Newf(1508, "host %s uses port 5044 (tcp/lumberjack), which is no longer supported. Use port 443 with protocol https")
+	ErrIngestionHostRemoved   = errors.Newf(1509, "host %s uses a legacy ingestion.* address, which is no longer supported. Use the phoenix.* host for your region")
+	ErrNoConnectionConfigured = errors.New(1510, "no connection configured")
+	ErrNoActiveConnection     = errors.New(1511, "no active connection")
```
jcollins-axway, line 15: We do not really use these error numbers anymore

pkg/traceability/backoff.go
sbolosan: We talked about this. I have it on my "whiteboard". I believe I can get to this as well 👍🏾

jcollins-axway: On that, do we need that http client in the package still? Backlog maybe but think we can clean it up.

sbolosan: Agree

pkg/traceability/failover.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"math/rand"
+	"strings"
+
+	"github.com/Axway/agent-sdk/pkg/event"
+)
+
+// failoverClient replaces libbeat's outputs.failoverClient.
+type failoverClient struct {
+	clients []NetworkClient
```
jcollins-axway, line 13: I think we will only have 1 Network Client going forward. However we would have to move some of this to the v7 agent.

pkg/traceability/tlsconfig.go
```diff
@@ -0,0 +1,62 @@
+package traceability
+
+import (
+	"github.com/Axway/agent-sdk/pkg/config"
+	"github.com/Axway/agent-sdk/pkg/util/log"
+)
+
+// unsupportedTLSField pairs a config key with whether it was set.
+type unsupportedTLSField struct {
```
jcollins-axway: 2 questions

Why a struct rather than a map[string]bool?

This is only used to store then log unsupported fields, seems overkill?

pkg/traceability/traceability.go
```diff
@@ -4,27 +4,17 @@ import (
...
 	"github.com/Axway/agent-sdk/pkg/util"
 	"github.com/Axway/agent-sdk/pkg/util/log"
-	"github.com/elastic/beats/v7/libbeat/beat"
-	"github.com/elastic/beats/v7/libbeat/common"
-	"github.com/elastic/beats/v7/libbeat/common/transport"
-	"github.com/elastic/beats/v7/libbeat/common/transport/tlscommon"
-	"github.com/elastic/beats/v7/libbeat/outputs"
-	"github.com/elastic/beats/v7/libbeat/paths"
-	"github.com/elastic/beats/v7/libbeat/publisher"
-	"golang.org/x/net/proxy"
```
jcollins-axway, line 27: Love the removals!

pkg/util/log/log.go
```diff
@@ -31,118 +25,61 @@ var networkTraceIgnoreHeaders = map[string]interface{}{
...
-	isLogP = true
-}
-
-// UnsetIsLogP -
-func UnsetIsLogP() {
-	isLogP = false
-}
-
 // Trace -
 func Trace(args ...interface{}) {
```
jcollins-axway, line 36: maybe we will be able to move all these in the future when we no validate all agents create a logger object and do not use a global logger.

Backlog item?

pkg/traceability/config.go
```diff
@@ -1,57 +1,99 @@
...
-	"github.com/elastic/beats/v7/libbeat/common"
-	"github.com/elastic/beats/v7/libbeat/common/cfgwarn"
-	"github.com/elastic/beats/v7/libbeat/common/transport/tlscommon"
+// legacy lumberjack/ingestion hostname prefixes no longer supported now that
+// single-entry routes HTTPS via phoenix.* hostnames (AOPS-4119)
+var removedIngestionHostPrefixes = []string{"ingestion.", "ingestion-http.", "ingestion-lumberjack."}
+
+// using the same env vars agents already use, e.g. pathHost -> TRACEABILITY_HOST.
+const (
+	pathHost              = "traceability.host"
```
jcollins-axway, line 27: Another backlog item here...

Maybe we can move all this to our common config package?

Even further, maybe we can start transitioning these to match our existing properties starting with central? Maybe not, but thought that hit me.

pkg/cmd/root.go
sbolosan: I took a look. NewCmd only still exists to wrap the libbeat-generated root command v7_traceability_agent builds (filebeat's own log-tailing). NewCmd can't go away until that's replaced, right? I'll discuss with @vivekschauhan  this week.

noted, backlog story - https://jira.axway.com/browse/APIGOV-33539

pkg/traceability/failover.go
sbolosan: I doubled check, v7_traceability_agent doesn't actually depend on the failoverClient. It wraps clients a second time in beats' own outputs.SuccessNet. But the risk here is every agent yaml defaults loadbalance: true, which bypasses failoverClient entirely, so it may already be unreachable in practice. So  it's a real configuration. So don't want to introduce a breaking change. Adding it to the backlog to look at properly.

noted, backlog story - https://jira.axway.com/browse/APIGOV-33539

pkg/traceability/tlsconfig.go
sbolosan: resolving both.  No reason not to use a map, and it is a bit of overkill

pkg/util/log/log.go
sbolosan: noted, backlog story - https://jira.axway.com/browse/APIGOV-33539

pkg/traceability/config.go
sbolosan: noted, backlog story - https://jira.axway.com/browse/APIGOV-33539

pkg/event/event.go
```diff
@@ -0,0 +1,29 @@
...
+
+// MapStr is a flat string-keyed metadata/fields map, replacing libbeat's common.MapStr.
+type MapStr map[string]interface{}
+
+// Event is a single event to publish, replacing libbeat's beat.Event/publisher.Event.
+type Event struct {
+	Timestamp time.Time
+	Meta      MapStr
+	Fields    MapStr
+	Private   interface{}
```
vivekschauhan, line 15: Probably we can remove this unused property

pkg/event/event.go
vivekschauhan, line 1: Should this be package be under traceability ?

go.mod
vivekschauhan: This should not be needed anymore as well

vivekschauhan: Just noticed that json.Marshal from that package being referenced in manager_test.go, should be encoding/json, so fix that we should be able to get rid of this dependency as well

Image

vivekschauhan (dismissed):

go.mod
sbolosan: right, fixing.

pkg/event/event.go
sbolosan: Private isn't actually unused — v7_traceability_agent relies on it to carry filebeat's file-read checkpoint through to the outgoing event, so filebeat only advances past a log line after it's actually published. Not visible from agent-sdk alone. I can add a comment here saying its still being used by v7 trace for now?

pkg/event/event.go
sbolosan: Good question. Checked this before answering. pkg/event is actually used by both pkg/traceability and pkg/transaction/pkg/transaction/metric, not just traceability, so it's more of a shared type between the two than something traceability-owned. Moving it under pkg/traceability would work without a cycle, but it would mean pkg/transaction, the side that builds events, has to reach into traceability's namespace just to get the basic Event type. Leaning toward keeping it standalone as its own small package, similar to how it replaced libbeat's core beat.Event rather than something output-specific. Open to other thoughts though.

pkg/transaction/eventreport.go
```diff
@@ -78,16 +78,16 @@ func (e *eventReport) GetEventTime() time.Time {
...
 	if e.metadata == nil {
-		e.metadata = common.MapStr{}
+		e.metadata = event.MapStr{}
 	}
 	return e.metadata
 }
 
-func (e *eventReport) GetFields() common.MapStr {
+func (e *eventReport) GetFields() event.MapStr {
 	if e.metadata == nil {
```
vivekschauhan: Not your change, but should this have been fields instead of metadata ?

pkg/traceability/config.go
```diff
@@ -84,27 +126,103 @@ func DefaultConfig() *Config {
...
+	props.AddIntProperty(pathPipelining, def.Pipelining, "Traceability client pipelining", properties.WithLowerLimitInt(0))
+	props.AddIntProperty(pathCompressionLevel, def.CompressionLevel, "Traceability client compression level",
+		properties.WithLowerLimitInt(0), properties.WithUpperLimitInt(9))
+	props.AddIntProperty(pathMaxRetries, def.MaxRetries, "Maximum number of retries for a failed traceability request",
+		properties.WithLowerLimitInt(-1))
+	props.AddStringProperty(pathSSLVerification, "", "TLS verification mode for the traceability client")
+	props.AddStringSliceProperty(pathSSLCipherSuites, []string{}, "Cipher suites allowed for the traceability client")
+	props.AddStringProperty(pathProxyURL, def.Proxy.URL, "SOCKS5 proxy URL for the traceability client")
+	props.AddBoolProperty(pathProxyLocalResolve, def.Proxy.LocalResolve, "Resolve names locally instead of on the SOCKS proxy server")
+	props.AddDurationProperty(pathBackoffInit, def.Backoff.Init, "Initial backoff duration for a failed traceability request", properties.WithLowerLimit(0))
```
vivekschauhan: Can you change the lower limit to 1, 0 could result in panic in backoffClient?

pkg/transaction/eventreport.go
sbolosan: good eyes.  fixing

pkg/config/centralconfig.go
```diff
@@ -66,26 +66,26 @@ var regionalSettingsMap = map[Region]regionalSettings{
...
 	},
 	AP: {
 		SingleURL:            "https://ingestion-ap-sg.platform.axway.com",
 		CentralURL:           "https://central.ap-sg.axway.com",
 		AuthURL:              "https://login.axway.com/auth",
 		PlatformURL:          "https://platform.axway.com",
-		TraceabilityHost:     "ingestion.visibility.ap-sg.axway.com:5044",
-		TraceabilityProtocol: "tcp",
+		TraceabilityHost:     "phoenix.visibility.ap-sg.axway.com:443",
+		TraceabilityProtocol: "https",
```
jcollins-axway: We can probably drop this attribute in the struct as the protocol will always be https.

pkg/traceability/config.go
jcollins-axway, line 27: not sure if you have it covered in the other followups, maybe we look to move this config to the config package.

pkg/util/dialer.go
```diff
@@ -22,6 +22,10 @@ const (
 	DefaultKeepAliveTimeout = 10 * time.Second
 )
 
+var dialerLogger = log.NewFieldLogger().
```
jcollins-axway: should our dialer logger be on the struct like our other loggers?

sbolosan: wasn't planning to leave this in.  was just in case the test V was doing failed again and could show us something.

pkg/traceability/config.go
sbolosan: Yeah, already have that on the backlog list I mentioned this morning. Moving traceability's Config/TLSConfig/ProxyConfig/Backoff into pkg/config to match everything else (CentralConfig, StatusConfig, etc.). Bigger lift than this round's cleanup.It's tracked and not forgotten.

pkg/util/dialer.go
sbolosan: moved it onto the dialer struct, set once in NewDialer and used as d.logger throughout, same pattern as Client.logger in traceability.go. thanks thanks

pkg/config/centralconfig.go
```diff
@@ -458,18 +453,7 @@ func (c *CentralConfiguration) GetTraceabilityHost() string {
 
 // GetTraceabilityProtocol - Returns the central traceability protocol
 func (c *CentralConfiguration) GetTraceabilityProtocol() string {
```
jcollins-axway: technically we can just always return "https" right?

jcollins-axway: can we kick it from the CentralConfig struct and the functions from the interface?

pkg/config/centralconfig.go
jcollins-axway: Are we able to drop this function altogether? Whatever is calling it should just know the protocol is always "https"

pkg/config/centralconfig.go
```diff
@@ -1122,7 +1122,7 @@ func (c *CentralConfiguration) setRegionBasedEnvironmentVars() {
 		"CENTRAL_URL":           c.GetURL,
 		"CENTRAL_AUTH_URL":      c.GetAuthConfig().GetTokenURL,
 		"TRACEABILITY_HOST":     c.GetTraceabilityHost,
-		"TRACEABILITY_PROTOCOL": c.GetTraceabilityProtocol,
+		"TRACEABILITY_PROTOCOL": c.traceabilityProtocol,
```
jcollins-axway: Can we completely remove this env var from existence?

pkg/config/centralconfig.go
sbolosan: right, frick

pkg/traceability/traceability.go
```diff
@@ -187,14 +187,9 @@ func NewClient(cfg *Config) ([]*Client, error) {
...
-	if isSingleEntry {
-		// only override with the expected protocol for a recognized single entry host. An
-		// unrecognized (e.g. customer-configured) single entry URL has no known protocol, so
-		// leave whatever was already configured alone rather than clobbering it with empty
-		if protocol := agent.GetCentralConfig().GetTraceabilityProtocol(); protocol != "" {
-			traceCfg.Protocol = protocol
-		}
+	if agent.GetCentralConfig().GetSingleURL() != "" {
+		// traceability ingestion is https-only
+		traceCfg.Protocol = "https"
```
jcollins-axway: Where ever traceCfg is initialized should have protocol hardcoded to "https" and nothing can change it.

pkg/traceability/traceability_test.go
```diff
@@ -521,8 +520,8 @@ func TestValidateProtocolPortUnrecognizedSingleURL(t *testing.T) {
 		},
 		"no single entry URL leaves the configured protocol alone": {
 			singleURL:        "",
-			configuredProto:  "https",
-			expectedProtocol: "https",
+			configuredProto:  "tcp",
+			expectedProtocol: "tcp",
```
jcollins-axway: Validations on protocol should no longer be needed as the value will be hardcoded.

vivekschauhan (dismissed):

pkg/config/centralconfig.go
sbolosan: Yep, dropped it entirely, it was essentially dead. validateProtocolPort() and FinishConfig's host fallback already force "https" literally wherever it matters, and DefaultConfig()'s own Protocol default is already "https" too.

pkg/config/centralconfig.go
sbolosan: removed

pkg/traceability/traceability.go
sbolosan: That'd remove a still-supported option though, right.  Unless i'm not seeing it, which is usually the case. ValidateCfg() only rejects "tcp",  "http" is still allowed by IsHTTPTransport(),  I'm guessing for non-TLS on-prem ingestion. Hardcoding "https" everywhere would break that for anyone using it. validateProtocolPort() already forces https for the one case that's always https (single-entry/SaaS)

sbolosan: Ok, updated sdk and updated v7 too. rawConfig no longer unpacks protocol at all, and toConfig() hardcodes Protocol: "https" directly, so nothing from yaml/env can touch it there either. Dropped the now-dead protocol: ${TRACEABILITY_PROTOCOL:...} line from both yamls and the README row.

vivekschauhan (dismissed):

jcollins-axway (dismissed):

vivekschauhan (approved):

PR #1095, opened 2026-08-27
APIGOV-33645 - update Platform API usages

jcollins-axway:
- Drop unused client funcs for getting user information
- Unify the platform api response structure across platform api calls
- Migrate to platform api for getting teams
  - Remove api call func to get team by name, use cache instead
  - Rename GetTeam -> GetTeams and create GetTeam to receive id for call to get single team
  - Limit fields requested for team call
  - Use cache for teams when possible as cache should update when team change made
- Update client mocks


pkg/apic/client.go
```diff
@@ -373,160 +371,124 @@ func (c *ServiceClient) sendServerRequest(url string, headers, query map[string]
...
+// getTeam - call the platform team api for a list of teams or a single team
+func (c *ServiceClient) getTeam(id string) (any, error) {
 	headers, err := c.createHeader()
 	if err != nil {
 		return nil, err
 	}
 
-	platformURL := fmt.Sprintf("%s/api/v1/org/%s", c.cfg.GetPlatformURL(), c.cfg.GetTenantID())
+	// set the fields we want from the team api
+	headers["fields"] = "guid,name,tags,default"
```
sbolosan: it looks like fields might be getting set on headers instead of the query params? and the request might be going out with an empty query map? 
Like GetEntitlements does it just above, since that one passes fields[] through as an actual query param.

pkg/agent/agent.go
```diff
@@ -437,9 +437,9 @@ func initEnvResources(cfg config.CentralConfig, client apic.Client) error {
 	cfg.GetCredentialConfig().SetExpirationDays(int(env.Policies.Credentials.Expiry.Period))
 
 	if cfg.GetTeamID() == "" {
-		team, err := client.GetCentralTeamByName(cfg.GetTeamName())
-		if err != nil {
-			return err
+		team := agent.cacheManager.GetTeamByName(cfg.GetTeamName())
```
sbolosan: is the cache populated at this point? RefreshTeamCache looks like it only runs later, during handleInitialization, after initEnvResources has already run

pkg/apic/mock/mockclient.go
```diff
@@ -216,9 +190,17 @@ func (m *Client) SetConfig(cfg corecfg.CentralConfig) {
...
+func (m *Client) GetTeam(id string) (*definitions.PlatformTeam, error) {
+	if m.GetTeamMock != nil {
+		return m.GetTeamMock(id)
+	}
+	return nil, nil
+}
+
+// GetTeams -
+func (m *Client) GetTeams() ([]definitions.PlatformTeam, error) {
 	if m.GetTeamMock != nil {
```
sbolosan: should this be m.GetTeamsMock? missing an 's'?

jcollins-axway: @sbolosan all of these comments have been handled, thank you!

sbolosan (dismissed):

dgghinea (approved):

vivekschauhan (approved):

PR #1097, opened 2026-09-02
APIGOV-33010 provisioning webhook dispatch

alrosca:


pkg/agent/handler/accessrequest.go
```diff
@@ -63,6 +77,9 @@ func (h *accessRequestHandler) ShouldHandle(ctx context.Context, event *proto.Ev
 	if action == proto.Event_SUBRESOURCEUPDATED && event.Metadata.GetSubresource() == defs.XAgentDetails {
 		return true
 	}
+	if action == proto.Event_SUBRESOURCEUPDATED && event.Metadata.GetSubresource() == defs.XWebhookDetails {
+		return true
+	}
```
jcollins-axway: Just a thought here, would like more opinions....

Considering we now can have handlers for the same type with different subresouces if event types specified, do we want to consider moving new actions like this to their own handler?

Another thought. Since the pattern is pretty equal, could all webhook subres handling be in the same handler amongst the provisioning resrouces?

Finally, should this check also validate that the webhook config is actually set on the handler?

jcollins-axway, line 83: Just a thought here, would like more opinions....

Considering we now can have handlers for the same type with different subresouces if event types specified, do we want to consider moving new actions like this to their own handler?

Another thought. Since the pattern is pretty equal, could all webhook subres handling be in the same handler amongst the provisioning resrouces?

Finally, should this check also validate that the webhook config is actually set on the handler?

pkg/agent/handler/accessrequest.go
```diff
@@ -43,6 +48,15 @@ func WithAccessRequestRetryCount(rc int) func(c *accessRequestHandler) {
 	}
 }
 
+// WithAccessRequestProvisioningWebhook configures the webhook the handler calls instead of its own
+// registered Provisioning implementation, when cfg.IsConfigured()
+func WithAccessRequestProvisioningWebhook(cfg config.ProvisioningWebhookEndpointConfig, client api.Client) func(c *accessRequestHandler) {
+	return func(c *accessRequestHandler) {
+		c.webhookCfg = cfg
+		c.webhookClient = client
```
jcollins-axway, line 56: Is this client a separate client from the one that is added to the handler already? Could the client already on the handler be updated to support both?

pkg/agent/handler/accessrequest.go
```diff
@@ -171,6 +199,11 @@ func (h *accessRequestHandler) Handle(ctx context.Context, meta *proto.EventMeta
 
 func (h *accessRequestHandler) onPending(ctx context.Context, ar *management.AccessRequest, mar *apiv1.ResourceInstance) *management.AccessRequest {
 	log := getLoggerFromContext(ctx)
+
+	if h.webhookCfg != nil && h.webhookCfg.IsConfigured() && webhookDispatchedFor(ar, webhookOperationProvision) {
```
jcollins-axway: does it makes sense to use a default webhook config on the handler that would result in IsConfigured returning false instead of having to check nil?

pkg/agent/handler/accessrequest.go
alrosca: Depending on the responsibilities we want the webhook to handle, a new handler for each resource might be necessary. Otherwise, a separate handler for the webhook subresource only might be a good idea

pkg/agent/handler/accessrequest.go
alrosca: I would not extend the handler client with webhook dispatch functionality because of the different purposes these two client have

pkg/agent/handler/accessrequest.go
alrosca: Fixed

pkg/agent/handler/credential.go
```diff
@@ -481,6 +656,41 @@ func (h *credentials) onUpdates(ctx context.Context, cred *management.Credential
 	return cred
 }
 
+// onWebhookUpdate dispatches pending credential update actions (suspend/rotate/enable) to the configured
+// webhook instead of calling this agent's own registered Provisioning implementation. Unlike onUpdates, it
+// does not call registerIDPClient on rotate - IDP client registration is the webhook's own responsibility
+// in webhook mode.
+func (h *credentials) onWebhookUpdate(ctx context.Context, cred *management.Credential, actions []prov.CredentialAction) error {
```
sbolosan, line 664: this func onWebhookUpdate looks like it only marks the update as "dispatched" after all actions in the batch succeed, not after each one. So if a credential has two pending actions and the second one fails, the first one (even though it already succeeded), look like gets resent on retry. For something like rotate, that means it could run twice, right?  I think onUpdates handles this correctly and avoids this by saving each action's result right after it completes.

pkg/config/provisioningwebhookconfig.go
```diff
@@ -0,0 +1,220 @@
...
+	props.AddStringProperty(path+".url", "", "URL of the webhook the agent calls instead of its own provisioning for "+resourceType+" events")
+	props.AddStringProperty(path+".headers", "", "Static headers to send with every "+resourceType+" provisioning webhook call")
+	props.AddStringProperty(path+".secret", "", "Bearer token for the "+resourceType+" provisioning webhook, used when authType is bearer")
+	props.AddStringProperty(path+".authType", string(ProvisioningWebhookAuthNone), "Authentication method for the "+resourceType+" provisioning webhook: none, basic, apiKey, bearer")
+	props.AddStringProperty(path+".username", "", "Username for "+resourceType+" provisioning webhook basic auth")
+	props.AddStringProperty(path+".password", "", "Password for "+resourceType+" provisioning webhook basic auth")
+	props.AddStringProperty(path+".apiKeyHeader", "", "Header name used to send the "+resourceType+" provisioning webhook API key")
+	props.AddStringProperty(path+".apiKeyValue", "", "API key value for the "+resourceType+" provisioning webhook")
+	props.AddIntProperty(path+".retryCount", 0, "Number of additional attempts to make if the "+resourceType+" provisioning webhook call fails")
```
sbolosan: Do we want to cap this? Right now, someone can set the webhook retry count to any number. Maybe like  provisioningRetryCount that gets capped at 3? It looks like each retry waits longer than the last (15s, then 30s, then 60s, and so on).  While it waits, it blocks, right? So if someone sets a big retry count, it could end up stuck holding up processing

pkg/agent/handler/webhook.go
```diff
@@ -0,0 +1,225 @@
+package handler
```
sbolosan, line 1: can we add tests for this file porfavor

pkg/agent/handler/webhook.go
```diff
@@ -0,0 +1,225 @@
...
+	v, _ := details[key].(string)
+	return v
+}
+
+// mirrorWebhookDetails copies the resource's x-webhook-details subresource into its x-agent-details, so
+// existing code (traceability lookups, etc.) that only knows how to read x-agent-details keeps working once
+// a provisioning webhook is configured - the webhook itself is scoped to write only x-webhook-details, not
+// x-agent-details directly. Returns false if there is nothing to mirror.
+func mirrorWebhookDetails(h subResourceCarrier) bool {
+	webhookDetails, ok := h.GetSubResource(defs.XWebhookDetails).(map[string]interface{})
```
sbolosan, line 84: maybe we need to skip the agent's own reserved keys (provisioningWebhookDispatched, status, message) when copying the webhook's data over, so that a customer's webhook can never accidentally overwrite the agent's internal "already dispatched" just by happening to send back a field with the same name?

pkg/agent/handler/webhook.go
jcollins-axway: done

pkg/config/provisioningwebhookconfig.go
jcollins-axway: Set upper to 5 and lower to 0

pkg/agent/handler/credential.go
jcollins-axway: So, lets talk about this a bit.

Yes, I do see your point that multiple actions not all being dispatched may cause some ruffles. But is that a bad thing in this case? We hit an error sending the second the receiver would have to handle the first event again. 

My concern would be tracking that each was sent and the keys we set that they were sent are never cleaned. So a new request in the same action would skip that action being sent again. We would need to future have an event or transaction id for this cred event to know that that is the one that was sent or not.

The question is, which path is better to have as a possible error path. Sending the same webhook request or never sending a webhook request?

pkg/agent/handler/webhook.go
jcollins-axway: done

pkg/agent/handler/credential.go
sbolosan: Good point.  I follow what you're putting down.  We good.

sbolosan (approved):

